PR #744, opened 2022-07-28
combined miopen db

cderb:
Combine perf db into find db (add params value to fdb table)
remove reference to perf db table
remove reference to perf config table
restructure import_db, export_db, worker_interface, evaluator, fin_eval, fin_builder to write perf db entries in context of find db table
remove extra perf db recording functions from fin_eval
add custom cache path to worker to avoid processes combatting over kdb file
add status output around fin results to provide more detailed errors for the job table
bug fixes


tuna/export_db.py
```diff
@@ -386,25 +392,29 @@ def insert_perf_db_sqlite(session, cnx, perf_db_entry, ins_cfg_id):
...
-        .filter(dbt.perf_db_table.valid == 1)\
-        .filter(dbt.perf_db_table.session == dbt.session.id)\
-        .filter(dbt.perf_db_table.miopen_config == dbt.perf_config_table.id)\
-        .filter(dbt.perf_config_table.config == dbt.config_table.id)
+    query = session.query(dbt.find_db_table,
+                          dbt.config_table, dbt.tensor_table)\
+        .filter(dbt.find_db_table.valid == 1)\
+        .filter(dbt.find_db_table.kernel_time != -1)\
+        .filter(dbt.find_db_table.params != '')\
+        .filter(dbt.find_db_table.session == dbt.session.id)\
```
alexandraBara, line 403: please add this as the first filtering param, followed by valid then the rest

tuna/fin_eval.py
```diff
@@ -195,9 +200,9 @@ def process_fdb_eval(self, fin_json, result_str='miopen_find_eval_result'):
           fdb_entry = obj
           fdb_entry.alg_lib = fdb_obj['algorithm']
           fdb_entry.kernel_time = fdb_obj['time']
-          # workspace
           fdb_entry.workspace_sz = fdb_obj['workspace']
           fdb_entry.session = self.dbt.session.id
+          fdb_entry.params = fdb_obj['params']
```
alexandraBara, line 205: is session being appended anywhere to fdb_entry like: fdb_entry.session = self.session.id?

tuna/import_db.py
```diff
@@ -260,8 +213,7 @@ def main():
   """main"""
   args = parse_args()
   args.table_cfg = ConvolutionConfig
-  args.table_perf_cfg = ConvPerfConfig
-  args.table_perf_db = ConvPerfDB
+  args.table_perf_db = ConvolutionFindDB
```
alexandraBara, line 216: can you please add the dbt structure to this script? initialize and then use its tables instead of importing specific miopen tables here.

alexandraBara (changes requested):

tuna/export_db.py
cderb: makes sense

tuna/fin_eval.py
cderb: line 204 looks like an equivalent.

tuna/import_db.py
```diff
@@ -212,8 +211,9 @@ def print_sqlite_rows(cnx, cfg_filter):
 def main():
   """main"""
   args = parse_args()
-  args.table_cfg = ConvolutionConfig
-  args.table_perf_db = ConvolutionFindDB
+  dbt = DBTables(session_id=args.session_id)
```
alexandraBara, line 214: i think i wrote DBTable to take in a null value for session_id as well. I think it would be more clear if you can:
dbt = DBTables(session_id=None)

alexandraBara (approved):

tuna/export_db.py
```diff
@@ -386,25 +392,28 @@ def insert_perf_db_sqlite(session, cnx, perf_db_entry, ins_cfg_id):
...
-        .filter(dbt.perf_db_table.valid == 1)\
-        .filter(dbt.perf_db_table.session == dbt.session.id)\
-        .filter(dbt.perf_db_table.miopen_config == dbt.perf_config_table.id)\
-        .filter(dbt.perf_config_table.config == dbt.config_table.id)
+    query = session.query(dbt.find_db_table,
+                          dbt.config_table, dbt.tensor_table)\
+        .filter(dbt.find_db_table.session == dbt.session.id)\
+        .filter(dbt.find_db_table.valid == 1)\
+        .filter(dbt.find_db_table.kernel_time != -1)\
+        .filter(dbt.find_db_table.params != '')\
```
JehandadKhan, line 403: [Question] What is the connotation for `params` to be not empty in the two cases when there already is an entry in the fdb and when there is none ?

cderb: There are of course tunable and non-tunable solvers
1. For non-tunable the fdb params field will always be an empty string.
2. For tunable the params field should be populated, but may be empty. Solvers not present in the perf db will return an empty string for params.

vars/utils.groovy
```diff
@@ -90,7 +90,7 @@ def finSolvers(){
 
 def finApplicability(){
     def tuna_docker = docker.build("ci-tuna:${branch_id}", "--build-arg FIN_TOKEN=${FIN_TOKEN} --build-arg BACKEND=HIPNOGPU .")
-    tuna_docker.inside("--network host  --dns 8.8.8.8 ") {
+    tuna_docker.inside("--network host  --dns 8.8.8.8 --device=/dev/kfd --device /dev/dri:/dev/dri:rw --volume /dev/dri:/dev/dri:rw --group-add video") {
```
JehandadKhan, line 93: (anchor)

JehandadKhan (approved):

PR #747, opened 2022-08-05
Multiprocess applicability testing

cderb:
With 32 processes, applicability for 100k configs now takes less than 5 min.
Added utility method for session retry taking a callback function + lamda invoker.

### DIFF
--- a/tuna/fin_class.py
+++ b/tuna/fin_class.py
@@ -344,7 +344,7 @@ class FinClass(WorkerInterface):
         #remove old applicability
         query = session.query(self.dbt.solver_app)\
           .filter(self.dbt.solver_app.session == self.session_id)\
-          .filter(self.dbt.solver_app_config == elem["input"]["config_tuna_id"])
+          .filter(self.dbt.solver_app.config == elem["input"]["config_tuna_id"])
         query.delete()
         if not elem["applicable_solvers"]:
           self.logger.warning("No applicable solvers for %s",
--- a/tuna/fin_class.py
+++ b/tuna/fin_class.py
@@ -473,4 +473,5 @@ class FinClass(WorkerInterface):
       self.applicability()
 
     self.multiproc = False
+    return False
 
--- a/tuna/fin_class.py
+++ b/tuna/fin_class.py
@@ -36,13 +36,12 @@ from sqlalchemy.exc import IntegrityError, InvalidRequestError  #pylint: disable
 from tuna.worker_interface import WorkerInterface
 from tuna.db_tables import connect_db
 from tuna.dbBase.sql_alchemy import DbSession
-#from tuna.utils.logger import setup_logger
 from tuna.metadata import get_solver_ids, FIN_CACHE
 from tuna.metadata import DOCKER_CMD, LOG_TIMEOUT, INVERS_DIR_MAP
 from tuna.fin_utils import compose_config_obj
 from tuna.tables import DBTables
 from tuna.config_type import ConfigType
-from tuna.helper import session_retry
+from tuna.utils.db_utility import session_retry
 
 
 class FinClass(WorkerInterface):
@@ -60,7 +59,6 @@ class FinClass(WorkerInterface):
     ])
     self.__dict__.update((key, None) for key in allowed_keys)
 
-    #self.logger = setup_logger('fin_class')
     connect_db()
     self.all_configs = []
     self.jobid_to_config = {}
@@ -221,12 +219,15 @@ class FinClass(WorkerInterface):
         ids = tuple([str(job_row.config) for job_row in rows])
         query = query.filter(self.dbt.config_table.id.in_(ids))
 
-      query = query.order_by(self.dbt.config_table.id.asc())
+      #order by id for splitting configs into blocks
+      query = query.order_by(self.dbt.config_table.id)
       rows = query.all()
 
-      extra = len(rows) % num_blk
-      start = idx * (len(rows) // num_blk)
+      #block_size = int(len(rows) // num_blk)
+      extra = len(rows) % num_blk  #leftover configs, don't divide evenly
+      start = idx * (len(rows) // num_blk)  #start of a process block
       end = (idx + 1) * (len(rows) // num_blk)
+      #distributing leftover configs to processes
       if idx < extra:
         start += idx
         end += 1 + idx
@@ -240,6 +241,7 @@ class FinClass(WorkerInterface):
       if start >= len(rows):
         return False
 
+      #copy the configs for this process
       if end >= len(rows):
         subarr = rows[start:]
       else:
@@ -250,6 +252,10 @@ class FinClass(WorkerInterface):
         if self.config_type == ConfigType.batch_norm:
           r_dict['direction'] = row.get_direction()
         self.all_configs.append(r_dict)
+
+    if self.all_configs is None:
+      return False
+
     return True
 
   def create_dumplist(self):
@@ -270,7 +276,6 @@ class FinClass(WorkerInterface):
 
       if not self.set_all_configs(idx, num_blk):
         return False
-      assert self.all_configs is not None
       return self.compose_fin_list()
 
     self.logger.error("Fin steps not recognized: %s", self.fin_steps)
@@ -337,10 +342,10 @@ class FinClass(WorkerInterface):
     for elem in json_in:
       if "applicable_solvers" in elem.keys():
         #remove old applicability
-        query = 'delete from {} where config={} and session={}'.format(
-            self.dbt.solver_app.__tablename__, elem["input"]["config_tuna_id"],
-            self.session_id)
-        session.execute(query)
+        query = session.query(self.dbt.solver_app)\
+          .filter(self.dbt.solver_app.session == self.session_id)\
+          .filter(self.dbt.solver_app_config == elem["input"]["config_tuna_id"])
+        query.delete()
         if not elem["applicable_solvers"]:
           self.logger.warning("No applicable solvers for %s",
                               elem["input"]["config_tuna_id"])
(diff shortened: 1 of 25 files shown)


alexandraBara:
Whats the speed up? How long does this take now?

tuna/fin_class.py
```diff
@@ -211,21 +208,44 @@ def applicability(self):
...
       if self.label:
-        query_cfgs = session.query(self.dbt.job_table)\
-            .filter(self.dbt.job_table.reason == self.label)
+        query_cfgs = session.query(self.dbt.config_tags_table)\
+            .filter(self.dbt.config_tags_table.tag == self.label)
         rows = query_cfgs.all()
         ids = tuple([str(job_row.config) for job_row in rows])
         query = query.filter(self.dbt.config_table.id.in_(ids))
 
+      query = query.order_by(self.dbt.config_table.id.asc())
```
alexandraBara, line 224: Is asc() needed here?

tuna/fin_class.py
```diff
@@ -211,21 +208,44 @@ def applicability(self):
...
+
+      extra = len(rows) % num_blk
+      start = idx * (len(rows) // num_blk)
+      end = (idx + 1) * (len(rows) // num_blk)
+      if idx < extra:
+        start += idx
+        end += 1 + idx
+      else:
+        start += extra
+        end += extra
```
alexandraBara, line 235: Can you explain what you are doing here? Maybe add some comments to the code

tuna/fin_class.py
```diff
@@ -242,7 +262,13 @@ def create_dumplist(self):
...
-      if not self.set_all_configs():
+      idx = 0
+      num_blk = 1
+      if self.multiproc:
+        idx = self.gpu_id
+        num_blk = self.num_procs.value
+
+      if not self.set_all_configs(idx, num_blk):
         return False
       assert self.all_configs is not None
```
alexandraBara, line 273: Possible to remove this assert and return false? I think we once decided we dont want asserts

tuna/fin_class.py
```diff
@@ -291,62 +317,68 @@ def dump_json(self, outfile, to_file=True):
...
 
     return ret
 
+  def insert_applicability(self, session, json_in):
+    """write applicability to sql"""
+    _, solver_id_map_h = get_solver_ids()
+    for elem in json_in:
+      if "applicable_solvers" in elem.keys():
+        #remove old applicability
+        query = 'delete from {} where config={} and session={}'.format(
```
alexandraBara, line 340: Can we turn this delete in SQLAlchemy format?

tuna/fin_class.py
```diff
@@ -291,62 +317,68 @@ def dump_json(self, outfile, to_file=True):
...
+                config=elem["input"]["config_tuna_id"],
+                session=self.session_id)
+            session.add(new_entry)
+          except KeyError:
+            self.logger.warning('Solver %s not found in solver table', solver)
+            self.logger.info("Please run 'go_fish.py --update_solver' first")
+            return False
+
+    self.logger.info('Commit bulk transaction, please wait')
+    session.commit()
```
alexandraBara, line 360: this needs to be in a try catch, possibly rollback if we are retrying

tuna/fin_class.py
```diff
@@ -291,62 +317,68 @@ def dump_json(self, outfile, to_file=True):
...
-              self.logger.warning('Solver %s not found in solver table', solver)
-              self.logger.info("Please run 'go_fish.py --update_solver' first")
-              return False
-      self.logger.info('Commit bulk transaction, please wait')
-      try:
-        session.commit()
-      except IntegrityError as err:
-        self.logger.warning("DB err occurred commiting bulk transaction %s",
-                            err)
+      #self.insert_applicability(session, json_in)
```
alexandraBara, line 371: remove

tuna/fin_class.py
```diff
@@ -33,32 +33,34 @@
...
+    super().__init__(**kwargs)
     allowed_keys = set([
         'fin_steps', 'arch_num_cu_list', 'local_file', 'fin_outfile',
         'fin_infile', 'machine', 'docker_name', 'version', 'config_type',
         'label', 'session_id'
     ])
     self.__dict__.update((key, None) for key in allowed_keys)
 
-    self.logger = setup_logger('fin_class')
+    #self.logger = setup_logger('fin_class')
```
alexandraBara, line 63: remove

tuna/fin_class.py
```diff
@@ -33,32 +33,34 @@
 from sqlalchemy import func as sqlalchemy_func
 from sqlalchemy.exc import IntegrityError, InvalidRequestError  #pylint: disable=wrong-import-order
 
+from tuna.worker_interface import WorkerInterface
 from tuna.db_tables import connect_db
 from tuna.dbBase.sql_alchemy import DbSession
-from tuna.utils.logger import setup_logger
+#from tuna.utils.logger import setup_logger
```
alexandraBara, line 39: remove?

tuna/helper.py
```diff
@@ -179,6 +180,24 @@ def mysqldb_overwrite_table(table, dict_list, filter_cols):
...
+      sleep(random.randint(1, 30))
+    except pymysql.err.OperationalError as error:
+      logger.warning('%s, maybe DB contention sleeping (%s)...', error, idx)
+      session.rollback()
+      sleep(random.randint(1, 30))
+
+  logger.error('All retries have failed.')
+  return None
+
+
```
alexandraBara, line 200: there is a db_helper.py in utils, maybe this would be better in there?

alexandraBara (changes requested):

tuna/fin_class.py
cderb: It doesn't matter which way actually. Is this a performance hit?

tuna/fin_class.py
cderb: This is splitting out the work into blocks for each process. So each process gets a non-overlapping set of configurations to test.

tuna/fin_class.py
alexandraBara: it can be, so if not needed i would remove.

tuna/fin_class.py
alexandraBara: OK, it would be useful to have this as a code comment.

tuna/fin_class.py
cderb: It is, take a look at the session_retry function

tuna/fin_class.py
cderb: I'll remove

tuna/helper.py
cderb: There are some other functions in helper with similar characteristics. Perhaps these should be migrated as well.

tuna/fin_class.py
cderb: sure

tuna/fin_class.py
```diff
@@ -211,25 +206,56 @@ def applicability(self):
...
           self.dbt.config_table).filter(self.dbt.config_table.valid == 1)
 
       if self.label:
-        query_cfgs = session.query(self.dbt.job_table)\
-            .filter(self.dbt.job_table.reason == self.label)
+        query_cfgs = session.query(self.dbt.config_tags_table)\
+            .filter(self.dbt.config_tags_table.tag == self.label)
         rows = query_cfgs.all()
         ids = tuple([str(job_row.config) for job_row in rows])
         query = query.filter(self.dbt.config_table.id.in_(ids))
```
JehandadKhan, line 220: Why not do this via join ?

alexandraBara (approved):

tuna/fin_class.py
```diff
@@ -291,62 +322,67 @@ def dump_json(self, outfile, to_file=True):
...
+    for elem in json_in:
+      if "applicable_solvers" in elem.keys():
+        #remove old applicability
+        query = session.query(self.dbt.solver_app)\
+          .filter(self.dbt.solver_app.session == self.session_id)\
+          .filter(self.dbt.solver_app.config == elem["input"]["config_tuna_id"])
+        query.delete()
+        if not elem["applicable_solvers"]:
+          self.logger.warning("No applicable solvers for %s",
+                              elem["input"]["config_tuna_id"])
```
JehandadKhan, line 351: With the versions in place we dont want to delete this ? Unless this respects the version information

tuna/utils/db_utility.py
```diff
@@ -46,6 +55,24 @@ def get_id_solvers():
   return id_solver_map_c, id_solver_map_h
 
 
+def session_retry(session, callback, actuator, logger=LOGGER):
```
JehandadKhan, line 58: I believe there are other places in our code, where we can make use of this function and do some cleanup, perhaps as a follow up PR

JehandadKhan: @reidkwja Can you please follow up on the comment above ?

JehandadKhan (changes requested):

tuna/utils/db_utility.py
cderb: @JehandadKhan I was thinking the same thing. But I'm waiting for 744, because of change conflicts there.

tuna/fin_class.py
cderb: This is reporting when the json returned contains an empty list for applicable_solvers. It shouldn't happen ever, but good to know.

tuna/fin_class.py
cderb: I'll address this.

tuna/fin_class.py
```diff
@@ -151,23 +153,16 @@ def get_fin_results(self):
     # pylint: disable=broad-except
     result = None
```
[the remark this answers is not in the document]
cderb: looks necessary for predictable behavior.

tuna/fin_class.py
JehandadKhan: This can also be accomplished using `limit` and `offset` in MySQL however, I refrained from suggesting that since that way would the MySQL session would be open much longer.

JehandadKhan (approved):

PR #753, opened 2022-08-11
Script arg cleaning & removal of old functionality.

cderb:
Remove outdate input arguments & clean up go_fish run script
Remove arch / num_cu from import_config & load_jobs
Remove old builder / evaluator, functionality replaced by fin_builder, fin_eval


tuna/go_fish.py
```diff
@@ -242,8 +204,7 @@ def parse_args():
   if args.fin_steps:
     has_fin = all(x in fin_session_steps for x in args.fin_steps)
 
-  if (args.update_applicability or has_fin or args.compile or
-      args.run_perf) and not args.session_id:
+  if (args.update_applicability or has_fin) and not args.session_id:
```
alexandraBara, line 207: are we using --compile and --eval for the new fin classes? if so, they need to be added here to make sure session_id is specified. If that is the case, i would review all these if's here and see if you can combine the checks in something shorter

tuna/go_fish.py
```diff
@@ -565,8 +492,6 @@ def compose_worker_list(res, args):
       LOGGER.info('launch mid %u, proc %u', machine.id, gpu_idx)
       if not launch_worker(gpu_idx, f_vals, worker_lst, args):
         break
-      if args.init_session:
```
alexandraBara, line 568: this is a short cut to only run init_session once, otherwise its ran as many times as there are GPU...which is fine, there is a duplicate check in the DB to avoid this, but its better if we stop it here. On a system with 8 gpus you would see the duplicate err 7 times.

tuna/import_configs.py
```diff
@@ -42,9 +42,8 @@
 
 def parse_args():
   """Parsing arguments"""
-  parser = setup_arg_parser(
-      'Import driver commands and perf db entries ',
-      [TunaArgs.ARCH, TunaArgs.NUM_CU, TunaArgs.VERSION, TunaArgs.CONFIG_TYPE])
+  parser = setup_arg_parser('Import driver commands and perf db entries ',
```
alexandraBara, line 45: With the idea that at some point this will become MIOpen code only, can we update this comment with: Import MIOpen Driver commands and MIOpen performance DB entries.

tuna/worker_interface.py
```diff
@@ -112,7 +113,6 @@ def __init__(self, **kwargs):
     self.compile_only = False
     self.docker_name = TUNA_DOCKER_NAME
     self.gpu = None
-    self.bin_cache = False
```
alexandraBara, line 115: i think we can also remove self.find_db = None. that is set in the self.dbt structure. Double check that no other tables are set here.

alexandraBara (changes requested):

tuna/go_fish.py
```diff
@@ -577,6 +502,7 @@ def main():
   args = parse_args()
 
   if args.list_solvers:
+    print_solvers()
```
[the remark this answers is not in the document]
cderb: Has the list_solvers option been deprecated?

tuna/go_fish.py
cderb: launch_worker returns false when init_session is true, so this is no longer necessary.

tuna/worker_interface.py
cderb: removing

tuna/go_fish.py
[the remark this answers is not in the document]
alexandraBara: Im not sure

alexandraBara (approved):

tests/test_abort_file.py
```diff
@@ -44,8 +44,8 @@ def add_job():
   find_configs = "SELECT count(*), tag FROM conv_config_tags WHERE tag='test_builder' GROUP BY tag"
 
   del_q = "DELETE FROM conv_job WHERE reason = 'tuna_pytest'"
-  ins_q = "INSERT INTO conv_job(config, state, solver, valid, reason, session) \
-        SELECT conv_config_tags.config, 'new', NULL, 1, 'tuna_pytest', 1 \
+  ins_q = "INSERT INTO conv_job(config, state, solver, valid, reason, session, fin_step) \
```
JehandadKhan, line 47: If we are changing this, we should move towards using using Alchemy ....

JehandadKhan (approved):

PR #757, opened 2022-08-16
Retry for fdb update

cderb:
Add a retry for sql interaction around updating find db entries.

### DIFF
--- a/tuna/fin_eval.py
+++ b/tuna/fin_eval.py
@@ -181,7 +181,7 @@ class FinEvaluator(WorkerInterface):
                                    is_temp=True)
 
   def update_fdb_eval_entry(self, session, fdb_obj):
-    """update fdb if individual fin json entry"""
+    """update fdb with individual fin json entry"""
     if fdb_obj['evaluated']:
       obj, _ = self.get_fdb_entry(session,
                                   self.solver_id_map[fdb_obj['solver_name']])
@@ -210,24 +210,21 @@ class FinEvaluator(WorkerInterface):
   def process_fdb_eval(self, fin_json, result_str='miopen_find_eval_result'):
     """process find db eval json results"""
     status = []
-    for fdb_obj in fin_json[result_str]:
-      self.logger.info('Processing object: %s', fdb_obj)
-      slv_stat = get_fin_slv_status(fdb_obj, 'evaluated')
-      status.append(slv_stat)
-
-      with DbSession() as session:
+    fdb_obj = None
+    with DbSession() as session:
+      for fdb_obj in fin_json[result_str]:
+        self.logger.info('Processing object: %s', fdb_obj)
+        slv_stat = get_fin_slv_status(fdb_obj, 'evaluated')
         callback = self.update_fdb_eval_entry
         actuator = lambda x: x(session, fdb_obj)
         #retry returns false on failure, callback return on success
         ret = session_retry(session, callback, actuator, self.logger)
+        if not ret:
+          self.logger.warning('FinEval: Unable to update Database')
+          slv_stat['success'] = False
+          slv_stat['result'] = 'FinEval: Unable to update Database'
 
-    if not ret:
-      self.logger.warning('FinEval: Unable to update Database')
-      status = [{
-          'solver': 'all',
-          'success': False,
-          'result': 'FinEval: Unable to update Database'
-      }]
+        status.append(slv_stat)
 
     return status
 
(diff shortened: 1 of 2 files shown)


tuna/fin_eval.py
```diff
@@ -179,45 +180,51 @@ def get_fin_input(self):
...
-              'success': False,
-              'result': 'FinEval: Unable to update Database: {}'.format(err)
-          }]
+    fdb_obj = None
+    with DbSession() as session:
+      for fdb_obj in fin_json[result_str]:
+        self.logger.info('Processing object: %s', fdb_obj)
+        slv_stat = get_fin_slv_status(fdb_obj, 'evaluated')
+        callback = self.update_fdb_eval_entry
+        actuator = lambda x: x(session, fdb_obj)
```
JehandadKhan, line 219: [Cosmetic] Perhaps these variables are not required, since the same may be passed in  directly.

tuna/fin_eval.py
```diff
@@ -179,45 +180,51 @@ def get_fin_input(self):
...
+        #retry returns false on failure, callback return on success
+        ret = session_retry(session, callback, actuator, self.logger)
```
JehandadKhan, line 221: Please note that the session object is outside the `for fdb_obj` loop and the retry is inside the loop, which means that if an exception happens half way through the loop there might be some entries which were committed and some that were not resulting in a hard to catch error.

@cderb  Can you please correct my understanding ?

cderb: @JehandadKhan There same session is used for all updates. For each object in the json results (fdb_key + solver) the session_retry function is called to attempt committing the entry. If all retries fail, the result status for that solver will be updated with the failure, and it will show in the job table result field. The session_retry function catches operational errors and loops. The function does not close the session.

tuna/fin_eval.py
cderb: changed

tuna/fin_eval.py
```diff
@@ -179,45 +180,50 @@ def get_fin_input(self):
...
-              'success': False,
-              'result': 'FinEval: Unable to update Database: {}'.format(err)
-          }]
+    fdb_obj = None
+    with DbSession() as session:
+      for fdb_obj in fin_json[result_str]:
+        self.logger.info('Processing object: %s', fdb_obj)
+        slv_stat = get_fin_slv_status(fdb_obj, 'evaluated')
+        #retry returns false on failure, callback return on success
+        ret = session_retry(session, self.update_fdb_eval_entry,
```
alexandraBara, line 219: if this fails and a DB err is thrown, where is that caught?

cderb: As of now, it is not. Only catching Op errors.

alexandraBara: I would do this: if an IntegrityError is caught, move on and log the err, if an OpErr is thrown retry? What do you think?

JehandadKhan (approved):

tuna/fin_eval.py
cderb: added

iq136boy (approved):

tuna/fin_eval.py
```diff
@@ -179,45 +180,50 @@ def get_fin_input(self):
...
+            'arch: %s, num_cu: %s, direction: %s',
+            self.config.id, self.solver_id_map[fdb_obj['solver_name']],
+            self.dbt.session.arch, self.dbt.session.num_cu, self.config.direction)
+        raise ValueError("Unable to query find db entry")
+      fdb_entry = obj
+      fdb_entry.alg_lib = fdb_obj['algorithm']
+      fdb_entry.kernel_time = fdb_obj['time']
+      fdb_entry.workspace_sz = fdb_obj['workspace']
+      fdb_entry.session = self.dbt.session.id
+      fdb_entry.params = fdb_obj['params']
```
iq136boy, line 200: I am current using 'staging' branch for tuning. In 'staging' branch, we suppose that the fdb table does not have column 'params'. My concern is that whether we could merge this branch to 'staging' without the 'params', so that we could do tuning without altering the fdb table.

iq136boy (dismissed):

tuna/fin_eval.py
cderb: The ixt-rack-15 db already has params in the fdb table. so this shouldn't cause an issue.

alexandraBara (approved):

PR #761, opened 2022-08-19
Import script update: create session for perf_db import

cderb:
import_db will create a session specific to the imported db.
User should now specify miopen_v + rocm_v
clean up uniqueness constraints on some tables.


tuna/import_db.py
```diff
@@ -211,6 +213,11 @@ def print_sqlite_rows(cnx, cfg_filter):
 def main():
   """main"""
   args = parse_args()
+
+  args.label = "imported perf db"
+  args.arch, args.num_cu = parse_pdb_filename(args.target_file)
```
alexandraBara, line 218: keep in mind this is going to need utest - if this already exists, it will need to be updated

cderb: utest added for function

tuna/session.py
```diff
@@ -67,14 +67,30 @@ class Session(BASE):
...
     """Add new session entry"""
-    self.arch = worker.machine.arch
-    self.num_cu = worker.machine.num_cu
-    self.rocm_v = worker.get_rocm_v()
-    self.miopen_v = worker.get_miopen_v()
     self.reason = args.label
-    if args.ticket:
-      self.ticket = args.ticket
     self.docker = args.docker_name
+    if hasattr(args, 'arch') and args.arch:
```
alexandraBara, line 72: why do we need to check hasattr & args.<>? is

tuna/miopen_tables.py
```diff
@@ -488,9 +482,8 @@ class BNGolden(BASE, GoldenMixin):
   """Golden table for batch norm"""
   __tablename__ = "bn_golden"
   __table_args__ = (UniqueConstraint("golden_miopen_v",
-                                     "find_db",
                                      "config",
-                                     "session",
```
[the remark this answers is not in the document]
cderb: session shouldn't be part of the golden table's unique constraints. There can be multiple session with overlapping information within arch/num_cu/config/solver, so it would confound the table.

tuna/session.py
cderb: In case args is passed to the function without those attributes.

alexandraBara:
@cderb  can you please update the code coverage %?

cderb:
> @cderb can you please update the code coverage %?

Pass on that looks like that is going to require some debugging. Not all of our pytests work with the same environment.

tuna/session.py
JehandadKhan: Wouldnt pylint complain about that ?

JehandadKhan (approved):

JehandadKhan:
@alexandraBara  Please review when convenient

alexandraBara (approved):

JehandadKhan:
@cderb  The PR has conflicts

PR #763, opened 2022-08-23
Default to local_machine for go_fish.py

cderb:
Make local_machine the default run mode for go_fish.


tuna/go_fish.py
```diff
@@ -71,11 +71,11 @@ def parse_args():
                       help=
                       'Session ID to be used as tuning tracker. ' \
                       'Allows to correlate DB results to tuning sessions')
-  parser.add_argument('--local_machine',
-                      dest='local_machine',
+  parser.add_argument('--remote_machine',
```
alexandraBara, line 74: where is this used?

iq136boy (approved):

tuna/go_fish.py
cderb: --local_machine was used everywhere so the inverse --romote_machine is used nowhere. But this allows use of the older functionality if desired.

alexandraBara: i see ok, if you think its worth keeping then we should keep it. Please document this somewhere

alexandraBara (approved):

PR #765, opened 2022-08-26
Check that session and env match

alexandraBara:
Added check for fin_builder/eval to ensure that session assumed rocm_v and miopen_v  match the runtime rocm/miopen_v


tuna/fin_builder.py
```diff
@@ -91,6 +91,13 @@ def process_pdb_compile(self, session, fin_json):
 
   def step(self):
     """Main functionality of the builder class. It picks up jobs in new state and compiles them"""
+
+    try:
+      self.check_env()
+    except ValueError as verr:
+      self.logger.error(verr)
+      return False
```
cderb, line 99: This will add extra file I/O for each job. Is there a clean way to check once before we launch the worker?

cderb (changes requested):

reidkwja (approved):
Looks good!

tuna/fin_builder.py
alexandraBara: ill investigate moving this call somewhere else

tuna/fin_eval.py
```diff
@@ -213,6 +213,12 @@ def update_fdb_eval_entry(self, session, fdb_obj):
 
   def process_fdb_eval(self, fin_json, result_str='miopen_find_eval_result'):
     """process find db eval json results"""
+    try:
+      self.check_env()
+    except ValueError as verr:
+      self.logger.error(verr)
+      return False
```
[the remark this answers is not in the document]
alexandraBara: this was a mistake, i moved it right at the top of step

tuna/fin_builder.py
cderb: Perhaps in the constructer for the Fin class objects? Or we could store state here to only check once during a run?

cderb (approved):

junliume (approved):

PR #769, opened 2022-08-31
coverage_fixes

reidkwja:


tests/covscripts/coverage_parse_attributes.py
```diff
@@ -24,18 +24,19 @@
 #
 ###############################################################################
 import json
-import sys
+import os
 
-sys.path.append("../tuna")
-sys.path.append("tuna")
+root_dir = os.path.dirname(
```
alexandraBara, line 29: @cderb is there a better way to get the root  dir in Tuna without having to do this?

cderb: Maybe by using relative path? "../../../"

reidkwja:
> 



> 

@alexandraBara 

Ahhh....For  that one, install coverage `pip install coverage` ... version == 6.4.2

alexandraBara (approved):
Paths have been fixed, LGTM

cderb (approved):

PR #770, opened 2022-09-07
Enforcing solver for load_job

alexandraBara:
Removed option to add jobs without specifying a solver.
Made only_appliable the default behavior, aka applicability table is always used to add jobs.
Added is_dynamic as an option filter.


tests/test_abort_file.py
```diff
@@ -38,30 +38,42 @@
...
+from utils import CfgImportArgs, LdJobArgs
+from tuna.tables import DBTables
+from tuna.import_configs import import_cfgs
+from tuna.db_tables import connect_db
+from load_job import test_tag_name as tag_name_test, add_jobs
 
 
 def test_abort():
+  #import configs
+  session = 1
```
cderb, line 50: Avoid hard coding session here. This will prevent the pytest from running stand-alone.

tests/test_find.py
```diff
@@ -35,26 +35,43 @@
...
+from utils import CfgImportArgs, LdJobArgs
+from tuna.db_tables import connect_db
+from tuna.tables import DBTables
+from tuna.import_configs import import_cfgs
+
+
+def add_job(arch, num_cu):
+
+  #import configs
+  session = 1
```
cderb, line 48: ditto, creation / existence of the session should be assured by the pytest.

cderb (changes requested):

tuna/load_job.py
```diff
@@ -80,11 +81,11 @@ def parse_args():
                              help='add jobs with only these solvers '\
                                '(can be a comma separated list)')
   parser.add_argument(
-      '-o',
-      '--only_applicable',
-      dest='only_app',
+      '-d',
+      '--only_dynamic',
+      dest='only_dynamic',
```
iq136boy, line 86: Make the 'load applicable solver' as default is a great idea which will avoid loading unnecessary solvers configs during the tuning process. The load_job script looks good to me.  @cderb, we need these changes in the staging branch for tuning on tuna_net2.

iq136boy (approved):

tests/test_find.py
alexandraBara: seems like this part of the code is not needed in this utest. Ill remove

cderb (approved):

tests/test_fin_builder.py
```diff
@@ -127,7 +112,12 @@ def test_fin_builder():
...
 
   with DbSession() as session:
-    get_jobs = f"SELECT count(*) from conv_job where session={args.session_id} and state in ('compiled');"
-    row = session.execute(get_jobs)
-    for row in res:
-      assert (row[0] == num_jobs)
+    valid_fin_err = session.query(dbt.job_table).filter(dbt.job_table.session==args.session_id)\
+                                         .filter(dbt.job_table.state=='errored')\
+                                         .filter(dbt.job_table.result.contains('%Find Compile: No results%'))\
+                                         .count()
```
JehandadKhan, line 118: Why subtract errored jobs? The test set we have is already very small ?

tests/test_find.py
```diff
@@ -85,5 +66,4 @@ def test_find():
 
   find_db = ConvolutionFindDB(**keys)
 
-  add_job('gfx908', 120)
```
JehandadKhan, line 88: I am assuming, the job is not required

tests/utils.py
```diff
@@ -103,6 +104,16 @@ class GoFishArgs():
   solver_id = None
 
 
+class DummyArgs(object):
+  """Dummy args object class to be used for testing"""
+
+  # pylint: disable=too-many-instance-attributes
```
JehandadKhan, line 110: The class has no attributes, is this still required ?

JehandadKhan:
@alexandraBara  Please note my comments on this PR and if need be please address them in a separate PR

PR #771, opened 2022-09-07
solver analytics integration

msaudulhassan:
Solver Analytics Integration

### DIFF
--- a/tuna/gen_fastdb.py
+++ b/tuna/gen_fastdb.py
@@ -266,27 +266,27 @@ def gen_fastdb(finddb, threshold=0, keep_fundamental=False, verbosity=0):
 
 def main():
   """ main """
-  default_in_filename = os.path.join(_DEFAULT_INPUT_DIR, 'finddb.pkl')
-  default_out_dirname = _DEFAULT_OUTPUT_DIR
+  default_in_file = os.path.join(_DEFAULT_INPUT_DIR, 'finddb.pkl')
+  default_out_dir = _DEFAULT_OUTPUT_DIR
 
   parser = argparse.ArgumentParser(description='Generate fastdb from finddb')
   parser.add_argument(
       '-i',
       '--in',
       type=str,
-      default=default_in_filename,
+      default=default_in_file,
       dest='input',
       help=
-      f'filename for the pickled finddb Pandas dataframe (current: {default_in_filename})'
+      f'filename for the pickled finddb Pandas dataframe (current: {default_in_file})'
   )
   parser.add_argument(
       '-o',
       '--out',
       type=str,
-      default=default_out_dirname,
+      default=default_out_dir,
       dest='output',
       help=
-      f'dir to output pickled fastdb pandas-dataframe to (current: {default_out_dirname})'
+      f'dir to output pickled fastdb pandas-dataframe to (current: {default_out_dir})'
   )
   parser.add_argument(
       '-t',
--- a/tuna/gen_fastdb.py
+++ b/tuna/gen_fastdb.py
@@ -1,0 +1,326 @@
+#!/usr/bin/env python3
+###############################################################################
+#
+# MIT License
+#
+# Copyright (c) 2022 Advanced Micro Devices, Inc.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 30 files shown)


tuna/gen_fastdb.py
```diff
@@ -0,0 +1,298 @@
...
+                    f"kernel_time: {row['kernel_time']}\n\n"
+
+      logging.log(duplicates_str, silent=True)
+
+    if strict:
+      raise AssertionError(
+          f'{num_duplicates} duplicate entries detected in {tag}finddb')
+    logging.warning(
+        f'{num_duplicates} duplicate entries delected in {tag}finddb')
+    issues.append(DUPLICATE_ENTRIES)
```
alexandraBara, line 104: if this is an issue, it should be covered by a unique_index in the DB and not something we should be checking for in python. We can discuss this offline on how to add/modify this constraint.

tuna/gen_fastdb.py
```diff
@@ -0,0 +1,298 @@
...
+
+  logging.log(f'checking {tag}finddb...', end_char='\r')
+
+  if len(finddb) == 0:
+    if strict:
+      raise AssertionError(f'{tag}finddb contains no entry')
+    logging.error(f'{tag}finddb contains no entry')
+    issues.append(NO_ENTRIES)
+
+  if (finddb['valid'] != 1).any():
```
alexandraBara, line 53: this should be covered by your db selection query, which should always select based on the filter valid=true. This makes sure that you never pick up invalid entries to look at. Then this check here, can go away

tuna/gen_fastdb.py
```diff
@@ -0,0 +1,298 @@
...
+  """ generate fastdb from finddb without thresholding/removing out any solvers """
+  if cols_with_conv_params is None:
+    cols_with_conv_params = _DEFAULT_COLS_WITH_CONV_PARAMS
+
+  num_duplicates = finddb.duplicated(subset=cols_with_conv_params +
+                                     ['solver']).sum()
+  if num_duplicates > 0:
+    logging.warning(f'resolved the {num_duplicates} duplicate (fdb_key, solver)' +\
+            f' pairs found in {tag}finddb by keeping the entries with fastest solver')
+  sorted_finddb = finddb.sort_values(cols_with_conv_params + ['kernel_time'])
```
alexandraBara, line 145: this sorting can and should be done in mysql, by using the func max() attribute in SQLAlchemy. Or, if you want to select all the entries including duplicates, you can sort by kernel_time asc or desc and then you can pick the fastest entry in constant time in python

tuna/gen_fastdb.py
```diff
@@ -0,0 +1,298 @@
...
+  def is_explicit_gemm(solver):
+    return is_substr('Gemm', solver) and not is_substr('ImplicitGemm', solver)
+
+  def is_naive(solver):
+    return is_substr('Naive', solver)
+
+  return is_explicit_gemm(solver) or is_naive(solver)
+
+
+# pylint: disable=too-many-locals, too-many-branches
```
alexandraBara, line 161: same here, please do not use these disable flags, we are using pylint to explicitly enforce addressing issues such as too-many-locals and too-many-branches

tuna/utils/function.py
```diff
@@ -0,0 +1,394 @@
+# pylint: disable=attribute-defined-outside-init
```
alexandraBara, line 1: please do not use this flag, set your attributes=None inside the init so pylint is happy. This is what we do in rest of Tuna

tuna/utils/helpers.py
```diff
@@ -0,0 +1,577 @@
...
+def compose(*fns):
+  """ compose(f, g) = g o f """
+  if len(fns) == 1:
+    return fns[0]
+  f = fns[0]  # pylint: disable=invalid-name
+  g = compose(*fns[1:])  # pylint: disable=invalid-name
+  return lambda x: g(f(x))
+
+
+# pylint: disable=function-redefined; need to fix this
```
alexandraBara, line 323: please fix this

alexandraBara (changes requested):
@msaudulhassan  please add copyrights to all the new files you have added.

How do you plan on adding tests for all the functionality you have added. We are currently at 10% coverage, i imagine with all this new code that does not have utest, that number will significantly decrease. We are not allowing for code that does not have code coverage to go in.
 
Please follow up on how you plan to add utest.

tuna/utils/function.py
msaudulhassan: defining an attribute as `None` has different semantics to not defining it at all (not just in theory, but the code is actually relying on this fact in some places)

tuna/utils/helpers.py
msaudulhassan: good catch!

tuna/gen_fastdb.py
msaudulhassan: added a flag called `invalid too`, so now you have to explicitly specify if you want invalid entries. The defaults won't trigger this warning

tuna/gen_fastdb.py
msaudulhassan: Don't really know MySQL, so I just converted conv_find_db table to a pandas dataframe in `gen_finddb.py`, so that all the scripts from there on process data frames

alexandraBara:
@JehandadKhan  do you think it would be a good idea to put all the solver analytics scripts in its designated folder? Trying to keep everything nice and tidy especially as we are moving towards separating library scripts into their respective folders.

msaudulhassan:
Created a separate repo for SolverAnalytics @ https://github.com/ROCmSoftwarePlatform/SolverAnalytics

PR #772, opened 2022-09-08
stable tuning branch

iq136boy:


requirements.txt
```diff
@@ -13,7 +13,7 @@ jedi==0.13.1
 lazy-object-proxy==1.3.1
 mccabe==0.6.1
 more-itertools==8.3.0
-numpy==1.22.0
+numpy==1.19.5
```
alexandraBara, line 16: @msaudulhassan  is this going to be an issue? downgrading numpy?

tuna/utils/db_utility.py
```diff
@@ -68,9 +68,13 @@ def session_retry(session, callback, actuator, logger=LOGGER):
...
       session.rollback()
       sleep(random.randint(1, 30))
+    except IntegrityError as error:
+      logger.error('Query failed: %s', error)
+      session.rollback()
+      return False
 
   logger.error('All retries have failed.')
-  return None
+  return False
```
alexandraBara, line 77: this function seems to always return False. Did you mean to return True at the end instead?

alexandraBara (changes requested):

requirements.txt
iq136boy: Hi Alex, it is still a draft now. Maybe I should merge develop to my branch, solve the conflict and then reopen the PR.

PR #776, opened 2022-09-16
Fix duplication of tensors

cderb:
Catch extra arguments to driver command. Filter 3d parameters inserted to 2d configs.


tuna/driver_conv.py
```diff
@@ -34,6 +35,7 @@
 from tuna.metadata import DIRECTION, DIR_MAP, CONV_SKIP_ARGS
 from tuna.parsing import get_fd_name, conv_arg_valid, get_fds_from_cmd
 
+LOGGER = setup_logger('driver_conv')
```
alexandraBara, line 38: I think it might be better to setup_logger in the driver_base.py class like this:
LOGGER = setup_logger(a.__class__.__name__)
then we dont have to worry about defining a name for it in each child class(conv, BN, fusions etc).
This line could then be removed

tuna/driver_conv.py
```diff
@@ -139,8 +141,13 @@ def config_set_defaults(self):
...
     """Set fds defaults"""
     for k, val in self.to_dict().items():
-      if val is None and k in defaults.keys():
-        setattr(self, k, defaults[k])
+      if k in defaults.keys():
+        if val is None:
+          setattr(self, k, defaults[k])
+        #for 2d configs filter out 3rd dimensional paramaters from unscrupulous users
+        elif self.spatial_dim != 3 and k.endswith('_d'):
+          setattr(self, k, defaults[k])
```
alexandraBara, line 149:
```suggestion
      if (k in defaults.keys() and val is None) or (self.spatial_dim != 3 and k.endswith('_d')):
          setattr(self, k, defaults[k])
```

alexandraBara (changes requested):

tuna/driver_conv.py
cderb: I think it's more readable this way.

tuna/driver_conv.py
cderb: You mean to setup the logger as a class variable like worker_interface?

alexandraBara: oh i just realized its not a class variable, for now, id rather not do that, lets just leave it as it.

alexandraBara (approved):

cderb:
merged with #782

PR #777, opened 2022-09-23
Rk pylint2

reidkwja:
Duplicate code restructuring + duplicate code argument remover from Pylint in Jenkin pipeline, allowing pylint to run.


alexandraBara (approved):
Nice

tuna/parsing.py
```diff
@@ -377,15 +377,22 @@ def arg_valid(arg, val):
...
   start = line.find('MIOpenDriver')
   if start == -1:
     raise ValueError(f"Invalid driver commmand line: '{line}'")
 
   line = line[start:]
+
+  return line
+
+
+def parse_driver_line(line):
```
cderb, line 392: This function appears to be unused. Maybe delete this instead of moving parse_line function here?

cderb (dismissed):

tuna/parsing.py
reidkwja: For the parse line, its used in the driver base

cderb: Referring to parse_driver_line which is defined in parsing, but never used anywhere.

alexandraBara: This is a bigger issue, and i think its beyond the scope of this PR. parse_driver_line in parsing is used in parsing.py but the functions that call it are not called anywhere. I will file an issue to remove this now dead code (since removal of Grafana functionality).

JehandadKhan (approved):

PR #778, opened 2022-09-23
Adding INT8 datatype

cderb:
remove BFP16, fdb key stores this as BF16 which exists
add INT8


alexandraBara:
@cderb do you have any driver cmds with int8? if so, can we add a couple to some of our utils/configs files, to make sure its included with some of our utests?

cderb:
@alexandraBara I'll key in the set from the tuning ticket.

iq136boy (approved):

alexandraBara (approved):

cderb:
merged with #782

PR #781, opened 2022-09-26
DB field rename: conv_mode -> mode

alexandraBara:
Renamed conv_mode to mode so it matches the Driver cmd fields.
This avoids having workarounds in the MITuna code to address the naming difference. 

Depends on: https://github.com/ROCmSoftwarePlatform/fin/pull/64


iq136boy (approved):

cderb:
How does this address conv_mode in metadata / parsing / driver_conv?

alexandraBara:
> How does this address conv_mode in metadata / parsing / driver_conv?

Those are all addressed in #783 
This PR only addressed the DB schema change through alembic.

alembic/versions/f3c8db790f9e_conv_mode_to_mode.py
```diff
@@ -0,0 +1,28 @@
...
+  op.alter_column('conv_config',
+                  'conv_mode',
+                  new_column_name='mode',
+                  existing_type=sa.VARCHAR(40),
+                  nullable=False,
+                  server_default="conv")
+
+
+def downgrade() -> None:
+  pass
```
JehandadKhan: Should we fill in the inverse of the above operation ?

JehandadKhan (changes requested):

alembic/versions/f3c8db790f9e_conv_mode_to_mode.py
alexandraBara: yes, ill update

JehandadKhan:
@cderb  Please review
@alexandraBara Please remove the `OnHold` label when ready

JehandadKhan (approved):

PR #783, opened 2022-10-04
New driver col in config tables

alexandraBara:
Added new driver col in config tables. This allows for easy query from Grafana.
Added the option to build a driver instance from a config db row.
Made direction an int everywhere
Renamed conv_mode -> mode to match the driver args from MIOpen. 
Note: left direction as is for now, might rename to forw later to match MIOpen.

Depends on:
https://github.com/ROCmSoftwarePlatform/fin/pull/64
https://github.com/ROCmSoftwarePlatform/MITuna/pull/781


Dockerfile
```diff
@@ -139,7 +139,7 @@ ARG FIN_TOKEN=
 RUN git clone https://$FIN_TOKEN:x-oauth-basic@github.com/ROCmSoftwarePlatform/Fin.git $FIN_DIR
 WORKDIR $FIN_DIR
 # Can be a branch or a SHA
-ARG FIN_BRANCH=260d9465d5ee71cc8dc7715ba583107ac4556f93
+ARG FIN_BRANCH=04508bcfe479310faafa4de8c7fe92e8cdf7ce5f
```
cderb, line 142: If you could link this off the develop branch when the linked Fin PR clears, that would be great.

tuna/driver_base.py
```diff
@@ -66,6 +73,18 @@ def construct_driver(self, line):
...
 
+  def construct_driver_from_db(self, db_obj):
+    """Takes a <>_config row and returns a driver cmd"""
+    LOGGER.info('Processing db_row: %s', db_obj.to_dict())
+    #common tensor among convolution and batch norm
+    self.decompose_input_t(db_obj)
+    if isinstance(db_obj, ConvolutionConfig):
+      self.parse_conv_row(db_obj)
+    elif isinstance(db_obj, BNConfig):
+      self.parse_bn_row(db_obj)
```
cderb: Perhaps it would be better to override a parse_row function in the driver child so that we don't leak the config types up into the generic class.

tuna/driver_base.py
```diff
@@ -31,19 +31,26 @@
...
-    if not self.construct_driver(line):
-      raise ValueError(f"Error creating Driver from line: '{line}'")
+  def __init__(self, line=None, db_obj=None):
+    if line:
+      if not self.construct_driver(line):
+        raise ValueError(f"Error creating Driver from line: '{line}'")
+    if db_obj:
+      if not self.construct_driver_from_db(db_obj):
+        raise ValueError(
+            f"Error creating Driver from db obj: '{db_obj.to_dict()}'")
```
cderb: What is the expected behavior for when line and db_obj are defined?

cderb (changes requested):

Dockerfile
iq136boy, line 142: I think we need a way to avoid the hardcode the fin commit hash.

iq136boy (changes requested):

Dockerfile
alexandraBara: hmm yes i will update this eventually but only when everything gets approved. I cannot push one PR without pushing all 3

alexandraBara: while fin is not a submodule of miopen and it often gets out of sync, we still need to do this. Pretty soon this will go away, we are about to open source Fin.

tuna/driver_bn.py
```diff
@@ -90,6 +90,10 @@ def compute_direction(self):
       raise ValueError("Can't import driver commmand line, \
           one and only one of forw or back must be set")
 
+  def parse_row(self, db_obj):
```
reidkwja, line 94: Is there a way to create unified function in driver base and perhaps inherit or import to bn and conv? This might trigger a pylint duplicate warning

reidkwja (changes requested):

tuna/driver_bn.py
alexandraBara: I believe that you need 2+rows to trigger a duplicate code warning, in this case its just 1 row(the function name). The other lines of code in this function are distinct in both child classes, so it does not trigger duplication. The pylint stage in the CI would have let us know by now. Technically this is the 'unified' function in the base class, overwritten in the child classes.

Dockerfile
JehandadKhan: Updated in develop!

alexandraBara: Not sure what you mean @JehandadKhan, my Tuna branch is pointing to my Fin branch, which is up to date with develop https://github.com/ROCmSoftwarePlatform/fin/pull/64

iq136boy (approved):

tuna/metadata.py
```diff
@@ -223,13 +227,22 @@
...
 }
 
+INVERS_CONV_TENSOR_PRECISION = {
+    'FP32': 'conv',
+    'FP16': 'convfp16',
+    'BF16': 'convbfp16',
+    'INT8': 'convint8'
+}
+
+INVERS_BN_TENSOR_PRECISION = {'FP32': 'bnorm', 'FP16': 'bnormfp16'}
```
cderb: These maps look like duplicates of PREC_TO_CMD at line 598.

tuna/driver_base.py
```diff
@@ -141,36 +159,34 @@ def compose_input_t(self):
...
     w_dict = self.compose_weight_t()
 
     return self.insert_tensor(w_dict)
 
-  def compose_weight_t(self):
-    """Build weight_tensor"""
-    w_dict = {}
-    w_dict['data_type'] = TENSOR_PRECISION[self.cmd]
-    w_dict['num_dims'] = self.num_dims
-
```
cderb, line 156: This is moved to driver_conv.py. Is this because it is convolution specific? Is compose_input_t shared by batchnorm?

cderb (changes requested):

reidkwja (approved):

tuna/driver_base.py
alexandraBara: yes

tuna/metadata.py
alexandraBara: you'r right, ill remove one of them

cderb (approved):

PR #795, opened 2022-10-17
add directions to conv configs

cderb:


alexandraBara (approved):

JehandadKhan (changes requested):
All these configs are now in the forward direction only, thus limiting the testing to only one direction. Can we split them between the different directions so that there is representation for everything?

JehandadKhan (approved):

PR #797, opened 2022-10-17
go_fish rewrite

alexandraBara:
Moving MIOpen out of go_fish and into its own lib file.
Expected go_fish cmd behavior:

- go_fish.py -h should print go_fish help (not MIOpen help)
- go_fish.py miopen --> should print MIOpen help
- go_fish.py miopen -h  --> should print MIOpen help
- go_fish.py  miopen --update_solvers --> typical update_solvers run

Anyother MIOpen cmd ran the same as before. MIOpen lib is default, no need to specify, but if you do specify the results will be the same.

Go fish was split into:

- go_fish.py (strictly start up script to launch library, get processes from lib and apply join())
- miopen/miopen_lib.py (MIOpen tuning lib, all miopen code)
- mituna_interface.py (Tuning lib interface - non MIOpen code)

All miopen scripts touch in this PR have been moved to miopen/ dir
Absolutely no functionality change applied, other than moving functions around.


tuna/libraries.py
```diff
@@ -0,0 +1,38 @@
...
+"""Module that encapsulates different liraries supported by Tuna"""
+from enum import Enum
+
+
+#pylint: disable=too-few-public-methods
+class Library(Enum):
+  """Enumerate supported libraries"""
+  # pylint: disable=invalid-name ; uppercasing would require modifying a lot of files
+  MIOPEN = 'miopen'
+
```
cderb, line 36: If Tuna is an extensible framework. Maybe it doesn't have complete knowledge of the libraries that can use it.

cderb (approved):

tuna/miopen/miopen_lib.py
```diff
@@ -0,0 +1,368 @@
...
+                        help=argparse.SUPPRESS,
+                        default=Library.MIOPEN,
+                        choices=Library,
+                        type=Library)
+
+    args = parser.parse_args()
+    if len(sys.argv) == 3 and ('--lib' in sys.argv or
+                               '-L' in sys.argv) and 'miopen' in sys.argv:
+      parser.print_help()
+      sys.exit(-1)
```
alexandraBara: @cderb  @JehandadKhan  do you have a better idea on how to handle this? I basically want it to return the -H when user only puts in: "go_fish.py -L miopen" with no other args.

tuna/libraries.py
alexandraBara: Hmm you have a point. Ill consider alternatives

tests/test_fin_builder.py
```diff
@@ -33,14 +33,15 @@
 this_path = os.path.dirname(__file__)
 
 from tuna.dbBase.sql_alchemy import DbSession
-from tuna.go_fish import load_machines, compose_worker_list
+from tuna.utils.miopen_utility import load_machines
```
JehandadKhan, line 36:
```suggestion
from tuna..miopen.utility import load_machines
```

tuna/go_fish.py
```diff
@@ -25,532 +25,52 @@
...
-      'miopen_find_compile', 'miopen_find_eval', 'miopen_perf_compile',
-      'miopen_perf_eval', 'get_applicability', 'find_compile', 'find_eval'
-  ]
-  has_fin = False
-  if args.fin_steps:
-    has_fin = all(x in fin_session_steps for x in args.fin_steps)
-
-  if (args.update_applicability or has_fin) and not args.session_id:
-    parser.error("session_id must be specified with this operation")
+  parser.add_argument('-L',
```
JehandadKhan: Can we just get rid of the `-L` and make it a positional arg

tuna/libraries.py
JehandadKhan: I am not sure I follow, @cderb  Are you saying a new lib might simply extend us and we wouldnt know about it ?

tuna/miopen/miopen_lib.py
```diff
@@ -0,0 +1,368 @@
...
+                       default=None,
+                       help='execute on each machine')
+
+    parser.add_argument('-L',
+                        '--lib',
+                        dest='lib',
+                        help=argparse.SUPPRESS,
+                        default=Library.MIOPEN,
+                        choices=Library,
+                        type=Library)
```
JehandadKhan: Perhaps a repetition?

tuna/miopen/miopen_lib.py
```diff
@@ -0,0 +1,368 @@
...
+        dest='find_mode',
+        type=int,
+        default=1,
+        help='Set the MIOPEN_FIND_MODE environment variable for MIOpen',
+        choices=[1, 3])
+    parser.add_argument('--remote_machine',
+                        dest='remote_machine',
+                        action='store_true',
+                        default=False,
+                        help='Run the process on a network machine')
```
JehandadKhan, line 75: Can we audit these args and just remove the ones we no longer use?

tuna/miopen/miopen_lib.py
```diff
@@ -0,0 +1,366 @@
...
+from tuna.fin_builder import FinBuilder
+from tuna.fin_eval import FinEvaluator
+from tuna.worker_interface import WorkerInterface
+from tuna.session import Session
+from tuna.utils.utility import get_env_vars, compose_f_vals, get_kwargs
+from tuna.utils.miopen_utility import load_machines
+from tuna.libraries import Library
+
+
+class MIOpen(MITunaInterface):
```
iq136boy, line 48: If someone wants to run other lib instead of miopen, we need to define/provide the lib interface, tables as in the miopen folder based on their applications. Something like we provide tuna APIs for other users to run their lib with tuna, right?

iq136boy (commented):
To my understanding, we want Tuna to support other libs/apps. The goal of the rewritings/changes is to make the 'integration' of other lib possible by providing the 'tuna APIs' as we did for miopen.

iq136boy (approved):

reidkwja (approved):
LGM

tuna/miopen/miopen_lib.py
alexandraBara: We will not provide that, another lib can extend our interfaces for instance db_tables.py (split i a different PR) and define its table there. Other things to define in a new library would be: a job, a worker etc all these things can be extended from the interfaces we provide, and lib specific implementations will be left up to the libs to define.

cderb (approved):
Moving forward I would be interested in seeing if we could accomplish this module linking without altering tuna. Meaning for a new module: a developer would flush out their tables & workers, pull the Tuna framework, and then no additional changes are required to Tuna. If we were still launching from go_fish, we would need to have a mechanism for linking to input parsing and a worker factory function. I wonder how to introduce this without updating go_fish.py or libraries.py with each new addition.

alexandraBara:
> Moving forward I would be interested in seeing if we could accomplish this module linking without altering tuna. Meaning for a new module: a developer would flush out their tables & workers, pull the Tuna framework, and then no additional changes are required to Tuna. If we were still launching from go_fish, we would need to have a mechanism for linking to input parsing and a worker factory function. I wonder how to introduce this without updating go_fish.py or libraries.py with each new addition.

Let me work on this, i have an idea

tuna/go_fish.py
alexandraBara: this was done

tuna/miopen/miopen_lib.py
alexandraBara: removed

tuna/miopen/miopen_lib.py
alexandraBara: ill take a look

PR #798, opened 2022-10-17
Adding Columns and Tables 

reidkwja:
The following PR adds two new tables to the database: 

- benchmark_table
- benchmark_perf_table


cderb (approved):

tuna/miopen_tables.py
```diff
@@ -518,6 +518,41 @@ class BNGolden(BASE, GoldenMixin):
...
+  model = Column(String(length=128), nullable=False)
+  batchsize = Column(Integer, nullable=False, server_default="32")
+  gpu_number = Column(Integer, nullable=True, server_default="1")  #nullable?
+  driver_cmd = Column(String(length=128), nullable=False)
+
+
+class BenchmarkPerfTable(BASE, GoldenMixin):
+  """benchmark table for performance parameters"""
+  __tablename__ = "benchmark_perf_table"
+  __table_args__ = (UniqueConstraint("kernel_time",
```
alexandraBara: we dont want kernel_time in the uq constraints, then we would have the same entry with multiple kernel times.

tuna/miopen_tables.py
```diff
@@ -518,6 +518,41 @@ class BNGolden(BASE, GoldenMixin):
...
+                                     "solver",
+                                     "workspace_sz",
```
alexandraBara: @JehandadKhan  should workspace_sz be part of the uq constraints? Im thinking no

tuna/miopen_tables.py
```diff
@@ -518,6 +518,41 @@ class BNGolden(BASE, GoldenMixin):
...
+                                     "rocm_v",
+                                     "miopen_v",
+                                     name="uq_idx"),)
+
+  config = Column(Integer, ForeignKey("benchmark_table.id"), nullable=False)
```
alexandraBara: Config needs to point to the actual config table. This is why your tables need to be renamed to conv_<table name> to make it apparent that we are referring to conv configs.
config = Column(Integer, ForeignKey("conv_config.id"), nullable=False)

tuna/miopen_tables.py
```diff
@@ -518,6 +518,41 @@ class BNGolden(BASE, GoldenMixin):
...
+  kernel_time = Column(DOUBLE, nullable=False, server_default="-1")
+  solver = Column(String(length=128), nullable=True, server_default="")
+  workspace_sz = Column(BigInteger, nullable=False)
+  rocm_v = Column(Integer, nullable=False)
+  miopen_v = Column(Integer, nullable=False)
+
```
alexandraBara: You benchmark_perf_table is also missing an FKey to the benchmark table. Something like this:
benchmark = Column(Integer, ForeignKey("benchmark_table.id"), nullable=False)

tuna/miopen_tables.py
```diff
@@ -518,6 +518,41 @@ class BNGolden(BASE, GoldenMixin):
...
+                                     name="uq_idx"),)
+
+  framework = Column(String(length=128), nullable=False)
+  model = Column(String(length=128), nullable=False)
+  batchsize = Column(Integer, nullable=False, server_default="32")
+  gpu_number = Column(Integer, nullable=True, server_default="1")  #nullable?
+  driver_cmd = Column(String(length=128), nullable=False)
+
+
+class BenchmarkPerfTable(BASE, GoldenMixin):
```
alexandraBara: Please rename this class as:
ConvBenchPerfTable or something like that

alexandraBara (changes requested):
Alternatively you can be a step ahead and make a BenchmarkPerfMixin and put everything but the config Fkey in there. I forsee us adding the same table for BN and then you will have to modify this anyway. Check any of the other Mixins that split based on the config col, one points to Conv and the other to BN in miopen_tables.py

tuna/miopen_tables.py
JehandadKhan: Workspace size cannot be part of the unique constraints

JehandadKhan: nor kernel_time for that matter

tuna/miopen_tables.py
```diff
@@ -518,6 +518,40 @@ class BNGolden(BASE, GoldenMixin):
   kernel_group = Column(Integer, nullable=True)
 
 
+class BenchmarkTable(BASE, GoldenMixin):
```
alexandraBara: Why are these classes inheriting GoldenMixin? This needs to be removed.

tuna/miopen_tables.py
```diff
@@ -518,6 +518,40 @@ class BNGolden(BASE, GoldenMixin):
...
+  __tablename__ = "conv_benchmark_perf_table"
+  __table_args__ = (UniqueConstraint("solver",
+                                     "rocm_v",
+                                     "miopen_v",
+                                     name="uq_idx"),)
+
+  config = Column(Integer, ForeignKey("conv_config.id"), nullable=False)
+  benchmark = Column(Integer, ForeignKey("benchmark_table.id"), nullable=False)
+  kernel_time = Column(DOUBLE, nullable=False, server_default="-1")
+  solver = Column(String(length=128), nullable=True, server_default="")
```
alexandraBara: this needs to be an FKey to the solver table (similar to the find_table solver column)

tuna/miopen_tables.py
```diff
@@ -518,6 +518,40 @@ class BNGolden(BASE, GoldenMixin):
...
+  workspace_sz = Column(BigInteger, nullable=False)
+  rocm_v = Column(Integer, nullable=False)
+  miopen_v = Column(Integer, nullable=False)
```
alexandraBara: these are not integers, they are strings. Please check session.py and copy paste the col declarations from there.

tuna/miopen_tables.py
```diff
@@ -518,6 +518,40 @@ class BNGolden(BASE, GoldenMixin):
...
+  model = Column(String(length=128), nullable=False)
+  batchsize = Column(Integer, nullable=False, server_default="32")
+  gpu_number = Column(Integer, nullable=True, server_default="1")  #nullable?
+  driver_cmd = Column(String(length=128), nullable=False)
+
+
+class ConvBenchPerfTable(BASE, GoldenMixin):
+  """benchmark table for performance parameters"""
+  __tablename__ = "conv_benchmark_perf_table"
+  __table_args__ = (UniqueConstraint("solver",
```
alexandraBara: the unique constraints need to be: config, benchmark, solver, rocm_v and miopen_v.

tuna/miopen_tables.py
```diff
@@ -518,6 +518,40 @@ class BNGolden(BASE, GoldenMixin):
...
+                                     "model",
+                                     "batchsize",
+                                     "gpu_number",
+                                     "driver_cmd",
+                                     name="uq_idx"),)
+
+  framework = Column(String(length=128), nullable=False)
+  model = Column(String(length=128), nullable=False)
+  batchsize = Column(Integer, nullable=False, server_default="32")
+  gpu_number = Column(Integer, nullable=True, server_default="1")  #nullable?
```
alexandraBara: remove the comment "#nullable" im ok with the default 1.

alexandraBara (changes requested):

tuna/miopen_tables.py
reidkwja: Should the onupdate/ondelete constraints remain there?

alexandraBara: yeah those are fine to stay

alexandraBara (approved):

PR #799, opened 2022-10-18
Moving MIOpen functionality out of Worker Interface

cderb:


tuna/worker_interface.py
```diff
@@ -89,54 +65,49 @@ def __init__(self, **kwargs):
 
     allowed_keys = set([
         'machine', 'gpu_id', 'num_procs', 'barred', 'bar_lock', 'envmt',
-        'reset_interval', 'fin_steps', 'fin_infile', 'fin_outfile', 'job_queue',
-        'queue_lock', 'label', 'fetch_state', 'docker_name', 'end_jobs',
-        'config_type', 'dynamic_solvers_only', 'session_id'
+        'reset_interval', 'fin_steps', 'job_queue', 'queue_lock', 'label',
```
alexandraBara, line 68: can we rename or move fin_steps somewhere else? its really miopen specific.

tuna/worker_interface.py
```diff
@@ -89,54 +65,49 @@ def __init__(self, **kwargs):
...
-    self.config_type = ConfigType.convolution if self.config_type is None else self.config_type
-    self.config_dict = None
     self.session_id = None
 
     self.__dict__.update(
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
 
+    #initialize tables
+    self.config_type = ConfigType.convolution if self.config_type is None else self.config_type
     self.dbt = DBTables(session_id=self.session_id,
```
alexandraBara, line 98: should we set DBTables in the child class? then we could move ConfigType into the miopen part as well. We could set self.dbt = self.set_db_tables() and then define this function in the miopen class.

alexandraBara (changes requested):

tuna/worker_interface.py
cderb: That's going to take more work, so it will be addressed in the next iteration. This also necessitates changes to the job queue.

tuna/worker_interface.py
cderb: Sure, that makes sense enough. I'll adapt that in the next PR.

alexandraBara (approved):

JehandadKhan (approved):

PR #801, opened 2022-10-20
Session Mixin

alexandraBara:


iq136boy (approved):

cderb (commented):
Would this now make session the miopen session? Migrating it to the miopen folder?
It seems to me like we would always have rocm_v, bar if we are tuning on another driver stack.

alexandraBara:
> Would this now make session the miopen session? Migrating it to the miopen folder? It seems to me like we would always have rocm_v, bar if we are tuning on another driver stack.

Ok so keep rocm_v in the Mixin and move the current session into miopen folder?

cderb (approved):

reidkwja (approved):

JehandadKhan:
@alexandraBara Please update the branch to address the conflicts

PR #803, opened 2022-10-24
Job Queue Refactor

cderb:
Split get_job functionality between worker_interface and fin_class. 
worker_interface - filter jobs by session characteristics and perform general job queueing. 
add result_queue for storing sql inserts
add compiled_pend and evaluated_pend states for completed jobs with uncommitted results.
correct reset_job_state logic to correctly identify and reset jobs in intermediate states.

fin_class - attaches config to the job query, implements job_queue_pop function to interpret queue items.
add result queue committing logic to bulk commit insert jobs


tuna/find_db.py
```diff
@@ -70,6 +71,8 @@ def get_query(self, sess, fdb_obj, session_id):
     if self.solver:
       query = query.filter(fdb_obj.solver == self.solver)
 
+    self.logger.info("fdb query %s-%s-%s-%s", session_id, self.config, self.solver, self.opencl)
```
msaudulhassan: Don't know why the linter didn't complain, but it usually wants f-strings

tuna/find_db.py
```diff
@@ -138,6 +141,8 @@ class ConvolutionFindDB(BASE, FindDBMixin):  #pylint: disable=too-many-instance-
   @orm.reconstructor
   def __init__(self, **kwargs):
     self.logger = kwargs['logger'] if 'logger' in kwargs else None  #pylint: disable=multiple-statements
```
msaudulhassan: `self.logger = kwargs["logger"] if "logger" in kwargs else setup_logger("find_db")`

tuna/miopen/fin_builder.py
```diff
@@ -85,18 +86,37 @@ def process_pdb_compile(self, session, fin_json):
...
-    except ValueError as verr:
-      self.logger.error(verr)
-      return False
+    if self.first_pass:
+      self.first_pass = False
+      try:
+        self.check_env()
+      except ValueError as verr:
+        self.logger.error(verr)
+        return False
```
JehandadKhan, line 115: Is the purpose here to only do the `check_env` once ? Dont we want this to happen before every job ?

tuna/miopen/fin_builder.py
```diff
@@ -85,18 +86,37 @@ def process_pdb_compile(self, session, fin_json):
...
     # pylint: enable=duplicate-code
 
     if not self.get_job("new", "compile_start", True):
+      while not self.result_queue_drain():
+        sleep(1)
```
JehandadKhan: Please implement a proper back off here. I believe the implementation exists in Tuna for a randomized backoff to prevent lockup

tuna/miopen/fin_builder.py
```diff
@@ -132,6 +152,10 @@ def step(self):
 
     if failed_job:
       self.set_job_state('errored', result=result_str)
+    elif self.pending:
+      self.set_job_state('compiled_pend', result=result_str)
```
JehandadKhan, line 157: I am not familiar with this state, can you elaborate what this represents ?

tuna/miopen/fin_class.py
```diff
@@ -78,10 +79,17 @@ def __init__(self, **kwargs):
     self.all_configs = []
     self.fin_list = []
     self.multiproc = False
+    self.pending = []
+    self.first_pass = True
 
     self.__dict__.update(
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
 
+    self.config_type = ConfigType.convolution if self.config_type is None else self.config_type
```
JehandadKhan, line 88: An unset config type should be an error if the higher layers will populate it ?

tuna/miopen/miopen_lib.py
```diff
@@ -334,7 +333,7 @@ def compose_worker_list(self, res, args):
           num_procs = int(env['slurm_cpus'])
         else:
           # JD: This sould be the responsibility of the machine class
-          num_procs = int(machine.get_num_cpus())
+          num_procs = int(machine.get_num_cpus() * .6)
```
JehandadKhan, line 336: What is the reason for scaling down here ?

tuna/worker_interface.py
```diff
@@ -177,78 +185,102 @@ def queue_end_reset(self):
...
+      return False
+    return True
 
-    if len(ids) != len(job_cfgs):
-      raise Exception(
-          f'Failed to load job queue. #ids: {len(ids)} - #job_cgfs: {len(job_cfgs)}'
-      )
-    for job, config in job_cfgs:
-      self.job_queue.put((job, config))
+  def get_job_table(self, job_tuple):
```
JehandadKhan: This method and the one below have very similar names, it would help if their names can be updated to avoid confusion

tuna/worker_interface.py
```diff
@@ -177,78 +185,102 @@ def queue_end_reset(self):
...
+    return job_tables
+
+  def refresh_query_objects(self, session, rows):
+    """refresh objects in query rows"""
+    for obj_tuple in rows:
+      try:
+        for entry in obj_tuple:
+          session.refresh(entry)
+      except TypeError:
+        session.refresh(obj_tuple)
```
JehandadKhan, line 232: Why not just refresh the entire tuple everytime ?

tuna/worker_interface.py
```diff
@@ -177,78 +185,102 @@ def queue_end_reset(self):
...
-              if not job_cfgs:
-                # we are done
-                self.logger.warning(
-                    'No %s jobs found, fin_step: %s, session %s', find_state,
-                    self.fin_steps, self.session_id)
-                if imply_end:
-                  self.logger.warning("set end")
-                  self.end_jobs.value = 1
+              job_query = self.compose_job_query(find_state, session)
+              job_rows = job_query.all()
```
JehandadKhan:
```suggestion
              job_rows = self.compose_job_query(find_state, session).all()
```

JehandadKhan (changes requested):
Can you help me understand how the `pending` list gets drained.

tuna/find_db.py
cderb: Sure, I can swap this to an f-string.

tuna/miopen/fin_builder.py
cderb: The check_env script executes the miopen_v check and rocm_v check and compares them against the session in sql. This is to save resources by limiting the check to once per worker. It's usually safe to assume that these details wouldn't change during a run.

tuna/miopen/fin_builder.py
cderb: sure, I can add the random here.

tuna/miopen/fin_builder.py
cderb: Pending is a list containing results that need to be inserted into the database. If it has members for the current job at the end of the step then they are added to the result queue, and the job is noted as compiled_pend. When the result queue is drained, the state will be updated to compiled.

tuna/miopen/fin_class.py
cderb: config_type is populated by FinClass, reference to config_type was removed from WorkerInterface. Default config type is convolution.

tuna/miopen/miopen_lib.py
cderb: I was encountering some deadlocking during testing. My theory was too many compilers. Scaling back the workers was helping.

tuna/worker_interface.py
cderb: I'll try and think of something.

tuna/worker_interface.py
cderb: I was under the impression that the db objects, had to be refreshed individually. If not then that would be a good solution.

tuna/find_db.py
cderb: Remembered that this is because of the logger function. The logger prefers this syntax.

cderb:
> Can you help me understand how the `pending` list gets drained.
The pending list is drained by the call to result_queu_drain(), defined in fin_builder / fin_eval. 
This would then call result_queue_commit(), defined in fin_class. result_queue_commit will iterate over each job in the result_queue, insert and commit the objects, and mark the job as completed.

tuna/find_db.py
msaudulhassan: oh I see

tuna/worker_interface.py
cderb: Refreshing the tuple doesn't appear to be possible.

tuna/miopen/fin_class.py
```diff
@@ -101,6 +105,47 @@ def chk_abort_file(self):
...
+  def base_job_query(self, session):
+    """select tables for job query"""
+    query = session.query(self.dbt.job_table, self.dbt.config_table)
+    return query
+
+  def compose_work_query(self, query):
+    """query for fin command and config"""
+    if self.fin_steps:
+      query = query.filter(
+          self.dbt.job_table.fin_step.like('%' + self.fin_steps[0] + '%'))
```
alexandraBara, line 122: Why use 'like' and not ==?

tuna/miopen/fin_class.py
```diff
@@ -622,6 +667,48 @@ def process_fdb_w_kernels(self,
 
     return status
 
+  def add_sql_objs(self, session, obj_list):
+    """add sql objects to the table"""
+    self.logger.info("adding pending #objects: %s", len(obj_list))
+    for sql_obj in obj_list:
+      session.add(sql_obj)
+    session.commit()
```
alexandraBara, line 675: could use a try catch in case we run into more db contention

tuna/miopen/fin_class.py
```diff
@@ -622,6 +667,48 @@ def process_fdb_w_kernels(self,
...
+      status = session_retry(session, self.add_sql_objs,
+                             lambda x: x(session, obj_list), self.logger)
+      if not status:
+        self.logger.error("Failed commit pending job %s", job_id)
+        continue
+
+      #set job states after successful commit
+      close_job()
+
+    self.job = this_job
```
alexandraBara, line 704: why are we saving the current job in this_job. And why are we setting self.job=job_dict['job'] in the loop?

tuna/miopen/miopen_tables.py
```diff
@@ -355,22 +357,25 @@ def session(self):
     """session key"""
     return Column(Integer, ForeignKey("session.id"), nullable=False)
 
+  reason = Column(String(length=60), nullable=False, server_default="")
   state = Column(Enum(JobEnum), nullable=False, server_default="new")
+  retries = Column(Integer, nullable=False, server_default="0")
+  result = Column(Text, nullable=True)
```
alexandraBara, line 363: needs and alembic change, here and other cols you added/removed bellow

tuna/worker_interface.py
```diff
@@ -270,43 +298,30 @@ def get_job(self, find_state, set_state, imply_end):
...
+          self.job = session.query(self.dbt.job_table)\
+              .filter(self.dbt.job_table.id == self.job.id).one()
+          self.job.state = state
+          if result:
+            self.job.result = result
+          if increment_retries:
+            self.job.retries += 1
+
+          if '_start' in state:
+            cache = '~/.cache/miopen_'
```
alexandraBara, line 317: hmm maybe we can set this location based on the library enum not hardcode the miopen?

alexandraBara (changes requested):

tuna/miopen/fin_class.py
cderb: not sure, this was some transplanted logic, maybe good time to change.

tuna/miopen/fin_class.py
cderb: this is executed inside a session_retry, see 695

tuna/miopen/fin_class.py
cderb: this_job saves the job currently associated with the worker. result_queue_commit iterates over inserts for jobs from any worker. Setting self.job so functions like set_job_state will function correctly. See close_job() implementation from fin_builder / fin_eval

tuna/miopen/miopen_tables.py
cderb: How to run alembic?

tuna/worker_interface.py
cderb: A fine idea, but not in scope for this pr.

tuna/miopen/fin_class.py
cderb: can't find jobs when using ==, leaving as is.

tuna/miopen/miopen_tables.py
cderb: Columns were moved in code, but no additions / deletions.

alexandraBara (approved):

JehandadKhan (approved):

PR #804, opened 2022-10-25
 unit tests : utility.py

technicalgrp89:


technicalgrp89:
@JehandadKhan @alexandraBara @cderb

tests/test_utility.py
```diff
@@ -0,0 +1,106 @@
...
+  arch_value = arch2targetid(arch)
+  assert (arch_value == f"{arch}:sram-ecc+:xnack-")
+
+
+def test_check_qts():
+
+  hostname = socket.gethostname()
+  m = Machine(hostname=hostname, local_machine=True)
+  retvalue = check_qts(hostname, logger=LOGGER)
+  print(retvalue)
```
alexandraBara: We dont need to print these retval's, they are just boolean, please remove everywhere.

alexandraBara (changes requested):

tests/test_utility.py
```diff
@@ -0,0 +1,106 @@
...
+from tuna.utils.utility import get_mmi_env_vars
+
+LOGGER = setup_logger('utility')
+
+
+def test_utility():
+  test_arch2targetid
+  test_check_qts
+  test_get_env_vars
+  test_get_mmi_env_vars
```
cderb: What does this function do? I'm pretty sure pytest is going to call every function in file that starts with 'test_'.

tests/test_utility.py
```diff
@@ -0,0 +1,106 @@
...
+  arch = 'gfx906'
+  arch_value = arch2targetid(arch)
+  assert (arch_value == f"{arch}:sram-ecc+:xnack-")
+
+
+def test_check_qts():
+
+  hostname = socket.gethostname()
+  m = Machine(hostname=hostname, local_machine=True)
+  retvalue = check_qts(hostname, logger=LOGGER)
```
cderb: The result of this test will change depending on the hostname of the machine. There's no guarantee the return from check_qts will be True or False.

tests/test_utility.py
```diff
@@ -0,0 +1,106 @@
...
+  print(retvalue)
+  assert (retvalue == False)
+
+  m = Machine(hostname=hostname, local_machine=True)
+  retvalue = check_qts(hostname, logger=LOGGER)
+  print(retvalue)
+  assert (retvalue == False)
```
cderb: this looks like the same test as 65-69

tests/test_utility.py
```diff
@@ -0,0 +1,106 @@
...
+
+  os.environ['user_name'] = 'xyz_123'
+  os.environ['user_password'] = 'xyz1234'
+  os.environ['db_hostname'] = 'xyz.test.com'
+  os.environ['db_name'] = 'testdb'
+
+  assert 'xyz_123' == (os.environ['user_name'])
+  assert 'xyz1234' == (os.environ['user_password'])
+  assert 'xyz.test.com' == (os.environ['db_hostname'])
+  assert 'testdb' == (os.environ['db_name'])
```
cderb: What does this test? Likely should contain a call to utils.utility.get_env_vars()

tests/test_utility.py
```diff
@@ -0,0 +1,106 @@
...
+
+def test_get_mmi_env_vars():
+
+  os.environ['gateway_ip'] = '10.100.00.000'
+  os.environ['gateway_port'] = '1234'
+  os.environ['gateway_user'] = 'xyz'
+
+  assert '10.100.00.000' == os.environ['gateway_ip']
+  assert '1234' == os.environ['gateway_port']
+  assert 'xyz' == os.environ['gateway_user']
```
cderb: Call to utils.utility.get_mmi_env_vars() ?

cderb (changes requested):

tests/test_utility.py
technicalgrp89: @cderb, Initially I had used same function call. Later, I changed to setting env variable values ,  while the execution of tests.
changing the *.env ,  environment variable values to dummy and test or set them using the os.environ while executing the unit test. In both we are comparing the env values. Inside utility.py to get the values os.environ[] is used.

cderb: I can see this function tests the os module from the python stl, but it doesn't test get_env_vars() from the tuna utility folder.

tests/test_utility.py
technicalgrp89: @cderb Instead of calling single tests indivisually, we can call/run multiple tests in a single file. Using 
pytest test_utility.py to execute all function tests or we can run single test test_check_qts.py

tests/test_utility.py
technicalgrp89: @cdrb what exactly, we should be cehcking here. As per the code trace, caller checks , if the check_qts returns Ture. Let me know

technicalgrp89:
@cderb @alexandraBara Addressed all review comments. Do let me know.

alexandraBara:
> @cderb @alexandraBara Addressed all review comments. Do let me know.

It seems like there are still issues brought up by @cderb  that have not been addressed.

tests/test_utility.py
technicalgrp89: print statements are removed

tests/test_utility.py
technicalgrp89: removed

tests/test_utility.py
technicalgrp89: get_env_vars() , replaced with the older code.

tests/test_utility.py
technicalgrp89: get_mmi_env_vars() replaced with the older code

technicalgrp89:
@alexandraBara , I have reolved except one review, waiting for Chris update.

tests/test_utility.py
cderb: Calling pytest test_utility.py will run every function in this file that is prefixed by 'test_'. Therefor this function is not necessary.

tests/test_utility.py
cderb: Let's assume that the current implementation is correct. 
If a hostname is in the QTS_LIST, which is a cache from the results of this function,- Then it returns that result.
If the hostname is an ip that starts with 192.168, then it returns True.
If the hostname is part of an entry in the machine table and has a local_ip then it returns true.

If you want to test against an entry in the machine table with certainty, you will have to add entries to the machine table in the associated test sql db.

The test here does not create a table entry for the hostname of the local machine. It also has no knowledge of the current state of that table. so there is no way to say deterministically that it will pass. So this can't be a correct test.

cderb (changes requested):

tests/test_utility.py
technicalgrp89: Ok. looks like, with current Tuna lib changes or usage, we are not using this functionality any more (depreicated). Let me remove this test.

JehandadKhan:
@technicalgrp89 As discussed, please add a line in Jenkinsfile to execute this test in the CI.

technicalgrp89:
@alexandraBara @cderb review comments , updated. let me know.

tests/test_utility.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_get_env_vars():
+
+  ENV_VARS = get_env_vars()
+
+  assert 'xyz_123' == (ENV_VARS['user_name'])
+  assert 'xyz1234' == (ENV_VARS['user_password'])
+  assert 'xyz.test.com' == (ENV_VARS['db_hostname'])
+  assert 'testdb' == (ENV_VARS['db_name'])
```
cderb, line 63: Test isn't self contained. Where are the environments set?

tests/test_utility.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_get_mmi_env_vars():
+
+  ENV_VARS = get_mmi_env_vars()
+  assert '10.100.00.000' == (ENV_VARS['gateway_ip'])
+  assert '1234' == (ENV_VARS['gateway_port'])
+  assert 'xyz' == (ENV_VARS['gateway_user'])
```
cderb: and here

cderb (changes requested):

tests/test_utility.py
technicalgrp89: I set tit using the .env file, source it , before stating the execution of tests.

tests/test_utility.py
technicalgrp89: @cderb same as above, Set uing the .env file, source it.

technicalgrp89: env variables are set inside test script.

tests/test_utility.py
technicalgrp89: set inside the test script

technicalgrp89:
@cderb code format is failing for file fin_utility.py , I have incorporated the review comments as well. Let me know.
@alexandraBara

technicalgrp89:
@cderb @alexandraBara CI checks are passing

alexandraBara (approved):

cderb (approved):

PR #807, opened 2022-10-28
New script to operate on benchmarks/framework/model tables

alexandraBara:
Added new script so we can update frameworks, models and use it to populate the benchmark table.


tuna/import_benchmark.py
```diff
@@ -0,0 +1,121 @@
...
+  with DbSession() as session:
+    for elem in FrameworkEnum:
+      try:
+        new_fmk = Framework(framework=elem)
+        session.add(new_fmk)
+        session.commit()
+        LOGGER.info('Added new framework %s', elem)
+      except IntegrityError as ierr:
+        LOGGER.warning(ierr)
+        continue
```
cderb, line 104: Should the order of the frameworks in the table be the same as that of the enum?

cderb (approved):

tuna/import_benchmark.py
alexandraBara: Yes, i think the above iteration guarantees that

iq136boy (approved):

reidkwja (approved):

tuna/import_benchmark.py
```diff
@@ -0,0 +1,121 @@
...
+
+from tuna.dbBase.sql_alchemy import DbSession
+from tuna.parse_args import TunaArgs, setup_arg_parser
+from tuna.utils.logger import setup_logger
+from tuna.miopen.benchmark import Framework, Model, FrameworkEnum, ModelEnum
+
+LOGGER = setup_logger('import_benchmarks')
+
+
+def parse_args():
```
JehandadKhan, line 38: Why is a new script required, why cant we add this to the current import_configs script ?

JehandadKhan (dismissed):

PR #811, opened 2022-11-07
Populating the benchmark table

alexandraBara:
Sample run:
./import_benchmark.py --add_benchmark -F Pytorch -m Alexnet -g 8 -f config.txt
where config.txt contains driver cmds associated with the Alexnet model.

Remove the conv_benchmark_perf table. This will be a view of conv_golden and benchmark join.


utils/configs/conv_configs_NCHW.txt
```diff
@@ -1,5 +1,5 @@
...
 
-./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --fil_layout NCHW --in_layout NCHW --out_layout NCHW
-./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --fil_layout NCHW --in_layout NCHW --out_layout NCHW
-./bin/MIOpenDriver conv -n 128 -c 256 -H 56 -W 56 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --fil_layout NCHW --in_layout NCHW --out_layout NCHW
+./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --fil_layout NCHW --in_layout NCHW --out_layout NCHW -F 2
+./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --fil_layout NCHW --in_layout NCHW --out_layout NCHW -F 4
+./bin/MIOpenDriver conv -n 128 -c 256 -H 56 -W 56 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --fil_layout NCHW --in_layout NCHW --out_layout NCHW -F 1
```
cderb: Direction appears to be stated 2x here.

utils/configs/conv_configs_NHWC.txt
```diff
@@ -1,9 +1,9 @@
...
+./bin/MIOpenDriver conv -n 32 -H 54 -W 54 -c 32 -k 32 --in_layout NHWC --out_layout NHWC --fil_layout NHWC -V 0 -F 1
 
-./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --fil_layout NHWC --in_layout NHWC --out_layout NHWC
-./bin/MIOpenDriver conv -n 128 -c 256 -H 14 -W 14 -k 256 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1 --fil_layout NHWC --in_layout NHWC --out_layout NHWC
-./bin/MIOpenDriver conv -n 128 -c 256 -H 56 -W 56 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --fil_layout NHWC --in_layout NHWC --out_layout NHWC
+./bin/MIOpenDriver conv -n 128 -c 1024 -H 14 -W 14 -k 256 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 1 -t 1 --fil_layout NHWC --in_layout NHWC --out_layout NHWC -F 2
+./bin/MIOpenDriver conv -n 128 -c 256 -H 14 -W 14 -k 256 -y 3 -x 3 -p 1 -q 1 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 4 -t 1 --fil_layout NHWC --in_layout NHWC --out_layout NHWC -F 4
+./bin/MIOpenDriver conv -n 128 -c 256 -H 56 -W 56 -k 64 -y 1 -x 1 -p 0 -q 0 -u 1 -v 1 -l 1 -j 1 -m conv -g 1 -F 2 -t 1 --fil_layout NHWC --in_layout NHWC --out_layout NHWC -F 1
```
cderb: and here

tuna/miopen/miopen_tables.py
```diff
@@ -530,44 +525,48 @@ class BNGolden(BASE, GoldenMixin):
...
-                                     "benchmark",
-                                     "solver",
-                                     "rocm_v",
-                                     "miopen_v",
+class BNBenchmark(BASE, BenchmarkMixin):
+  """benchmark table for framework and model parameters"""
+  __tablename__ = "bn_benchmark"
+  __table_args__ = (UniqueConstraint("framework",
+                                     "model",
+                                     "batchsize",
```
cderb, line 564: batchsize is a component of the config

tuna/miopen/miopen_tables.py
```diff
@@ -530,44 +525,48 @@ class BNGolden(BASE, GoldenMixin):
...
+                                     "gpu_number",
```
cderb, line 565: What does gpu_number describe? Do we use multiple gpus for certain models?

cderb (changes requested):

tuna/miopen/miopen_tables.py
alexandraBara: yes you are right, let me check with @JehandadKhan if there is a reason why we want this to be here as well.

tuna/miopen/miopen_tables.py
alexandraBara: I think its because we can run benchmarks on machines with different number of GPUs and that might affect the overall benchmark performance. @JehandadKhan want to chime in?

tuna/miopen/miopen_tables.py
JehandadKhan: Can we use this field to collect configs for different benchmark runs of the same benchmark with different batch sizes ?

tuna/miopen/miopen_tables.py
JehandadKhan: I think it would help if we rename it to `gpu_count` ? Since multi-GPU benchmarks have a number of GPUs associated with them. MIOpen still runs each config on a single GPU though

tuna/miopen/miopen_tables.py
alexandraBara: u mean use this to overwrite the batchsize for the same config? That is not how its currently used, but yes i can update it to overwrite it

tuna/miopen/miopen_tables.py
alexandraBara: OK, should we default it to 1 if not specified?

JehandadKhan: I suggest make it explicit, so that when we are importing benchmarks, we have this info explicitly added rather than assumed.

tuna/miopen/miopen_tables.py
JehandadKhan: I think I recall our discussion, this is actually an input to the benchmark so this should come from the command line. The batchsize in the config might be different from the batchsize of the benchmark.

alexandraBara: Im going to resolve this, because the batchsize here refers to the models batchsize and needs to be kept separate from the configs batchsize

tuna/import_benchmark.py
```diff
@@ -25,33 +25,78 @@
...
+                      required=False,
+                      help='Specify driver cmd')
+  parser.add_argument('-g',
+                      '--gpu_count',
+                      dest='gpu_count',
+                      type=int,
+                      default=None,
+                      required=False,
+                      help='Specify number of gpus the benchmark runs on')
+  parser.add_argument('-f',
```
JehandadKhan, line 95: Is `batchsize` missing ?

JehandadKhan (changes requested):

tuna/import_benchmark.py
alexandraBara: Oh good catch, i must have deleted it. Re-adding

JehandadKhan (approved):

reidkwja (approved):
Reviewed Import_benchmark/testing...It makes sense to me so far!

cderb (approved):

PR #812, opened 2022-11-07
Analytics function for fdb to pdb comparison

cderb:
fix to driver_conv so that constructing with fdb key does not add fields
add comparison operator do driver_conv
correct parsing and metadata so that layout is properly updated for fdb key parsing
add analyze_fdb script for testing presence of pdb entries given a fdb record.


tuna/driver_conv.py
```diff
@@ -270,3 +270,12 @@ def decompose_weight_t(self, db_obj):
...
     self.cmd = PREC_TO_CMD[ConfigType.convolution][data_type]
+
+  def __eq__(self, other):
+    if not isinstance(other, DriverConvolution):
+      return NotImplemented
+
+    if self.vars() == other.vars():
+      return True
+
+    return False
```
alexandraBara: why does this need to overwrite the driver_base eq function? It seems to check the same things.

cderb: I see, perhaps this isn't necessary.

alexandraBara (changes requested):
Do we have unit tests for the analyze_fdb file? If not, we need to add some.
How about some tests for the updates parsing functions?

alexandraBara (approved):

JehandadKhan:
@cderb  Please help with the CI run

JehandadKhan (approved):

PR #815, opened 2022-11-09
Tuna-unit tests for merge_db.py

technicalgrp89:
Following are the unit test cases added , into test_merge_db_functions.py
  test_parse_jobline
  test_parse_text_fdb_name
  test_parse_text_pdb_name
  test_load_master_list
  test_best_solver
  test_target_merge
  test_update_master_list
  test_write_merge_results
  test_merge_text_file
  test_get_sqlite_table
  test_get_sqlite_row
  test_get_sqlite_data
  test_get_file_list
/utils/coverage_files/coverage_percentage.txt is updated with latest coverage results.


technicalgrp89:
@JehandadKhan @cderb @alexandraBara , Need a code review, Let meknow.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+def test_parse_jobline():
+  data = """1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F=miopenConvolutionFwdAlgoImplicitGEMM:
+        ConvHipImplicitGemmForwardV4R4Xdlops_Padded_Gemm,
+        0.02352,0,miopenConvolutionFwdAlgoImplicitGEMM,not used;
+        miopenConvolutionFwdAlgoWinograd:ConvBinWinogradRxSf2x3g1,0.03856,0,miopenConvolutionFwdAlgoWinograd,
+        not used;miopenConvolutionFwdAlgoDirect:ConvOclDirectFwdGen,0.0536,0,miopenConvolutionFwdAlgoDirect,not used;
+        miopenConvolutionFwdAlgoGEMM:GemmFwdRest,0.05712,2749200,miopenConvolutionFwdAlgoGEMM,not used"""
+
+  key, vals = parse_jobline(data)
+  assert key, vals
```
alexandraBara: I think a better assertion would be to check for actual expected value like:
assert key = "someExpectedKey" and same for vals

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+  assert (arch == "gfx1030")
+  assert (num_cu == 36)
+  assert (final_file == "/data/tests/gfx1030_36.cd.pdb.txt")
+  assert (copy_files == [])
+
+
+def test_load_master_list():
+  master_file = '/data/tests/gfx803_36.HIP.fdb.txt'
+  master_list = load_master_list(master_file)
+  if master_list == {}:
```
alexandraBara, line 120: there should be some values in master_list correct? It should check for something other than empty or not

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+
+
+def test_best_solver():
+  #valid sample input test
+  test_value = ({
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvMlirIgemmFwdXdlops,0.03776,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  })
+
+  new_solver, new_time = best_solver(test_value)
```
alexandraBara: i think this test would be better if there were 2 solvers in the test_value entries and then the assertion would be checking for the fastest one.
Same for the rest of the best_solver calls bellow

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+      }
+  }
+
+  key = '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'
+  vals = {
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvMlirIgemmFwdXdlops,0.03776,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  }
+  keep_keys = False
+  target_merge(master_list, key, vals, keep_keys)
```
alexandraBara, line 207: does this function return anything? there should be an assert after the target_merge() call

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+          'miopenConvolutionFwdAlgoGEMM':
+              'GemmFwdRest,0.06096,414720,miopenConvolutionFwdAlgoGEMM,not used'
+      }
+  }
+
+  key = '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'
+  local_paths = ['/data/tests/../utils/test_files/usr_gfx90a68.HIP.fdb.txt']
+  mids = [-1]
+  keep_keys = False
+  update_master_list(master_list, local_paths, mids, keep_keys)
```
alexandraBara, line 409: same here, there needs to be a check for return value and some checks that the final data was actually updated correctly. That the best value was chosen etc

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+              'ConvBinWinogradRxSf2x3g1,0.03104,0,miopenConvolutionFwdAlgoWinograd,not used',
+          'miopenConvolutionFwdAlgoGEMM':
+              'GemmFwdRest,0.06096,414720,miopenConvolutionFwdAlgoGEMM,not used'
+      }
+  }
+
+  final_file = '/data/tests/old_gfx90a68.HIP.fdb.txt'
+  copy_files = []
+
+  write_merge_results(master_list, final_file, copy_files)
```
alexandraBara, line 472: this is also missing an assert. We need to check for expected values

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+
+  result_file = merge_text_file(master_file, copy_only, keep_keys, target_file)
+
+  if (os.stat(result_file)):
+    assert (True)
+  else:
+    assert (False)
+
+  #targetr_file set to None
+  result_file = merge_text_file(master_file=None,
```
alexandraBara: missing asserts

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+
+  if not perf_rows:
+    assert (False)
+  else:
+    assert (True)
+
+  if not perf_cols:
+    assert (False)
+  else:
+    assert (True)
```
alexandraBara: can probably merge this with:
if not perf_rows and perf_cols:
assert (False)
  else:
    assert (True)

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+
+  local_path = '/data/utils/test_files/test_gfx90678.db'
+  cnx_from = sqlite3.connect(local_path)
+
+  perf_rows, perf_cols = get_sqlite_table(cnx_from, 'perf_db')
+  for row in perf_rows:
+    perf = dict(zip(perf_cols, row))
+
+    cfg_row, cfg_cols = get_sqlite_row(cnx_from, 'config', perf['config'])
+    print(type(cfg_row), type(cfg_cols))
```
alexandraBara: not printouts pls

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+      'bias': 0,
+      'group_count': 1
+  }
+
+  cnx_from = sqlite3.connect(local_path)
+  cnx_to = sqlite3.connect(final_file)
+
+  res, col = get_sqlite_data(cnx_to, 'config', prune_cfg_dims(cfg))
+
+  print(type(res), res)
```
alexandraBara: not printouts anywhere pls

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+  print(type(col), col)
+
+  if not res:
+    assert (False)
+  else:
+    assert (True)
+
+
+def test_get_file_list():
+  print('test - get file list')
```
alexandraBara: this...seems to be missing something?

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+  test_load_master_list
+  test_best_solver
+  test_target_merge
+  test_update_master_list
+  test_write_merge_results
+  test_merge_text_file
+  test_get_sqlite_table
+  test_get_sqlite_row
+  test_get_sqlite_data
+  test_get_file_list
```
alexandraBara: this is not necessary, pytest will call every function in this script that starts with 'test_<>'. This can be removed

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+from tuna.utils.merge_db import write_merge_results
+from tuna.utils.merge_db import merge_text_file
+from tuna.utils.merge_db import merge_sqlite_pdb
+from tuna.utils.merge_db import merge_sqlite_bin_cache
+from tuna.utils.merge_db import merge_sqlite
+from tuna.utils.merge_db import get_file_list
+from tuna.utils.merge_db import get_sqlite_table
+from tuna.utils.merge_db import get_sqlite_row
+from tuna.helper import prune_cfg_dims
+from tuna.utils.merge_db import get_sqlite_data
```
alexandraBara: please condense these

alexandraBara (changes requested):

tests/test_merge_db_functions.py
technicalgrp89: assert are checked for key and vals

tests/test_merge_db_functions.py
technicalgrp89: added checks for the content of master list.

alexandraBara (changes requested):
I dont see the new utest file being called in the groovy file

tests/test_merge_db_functions.py
cderb: in this case:
assert key == fdb_key

for algorithm in data:
    assert algorithm in vals

for alg_data in data:
    assert alg_data[algorithm] == vals[algorithm]

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+  print(type(new_time), new_time)
+  if (len(new_solver) == 0):
+    assert (False)
+  else:
+    assert (True)
+
+  if (float(new_time)):
+    assert (True)
+  else:
+    assert (False)
```
cderb: don't need prints.
the solver string and time should be known based on the test inputs.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+  test_value = ({
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          ' ,0.00000,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  })
+
+  new_solver, new_time = best_solver(test_value)
+  if (not (new_solver and new_solver.isspace())):
+    assert (True)
+  else:
+    assert (False)
```
cderb: Is this a test for an improperly formatted fdb entry? Why the solver name in particular?

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+
+  #Sample as None and time 0.0
+
+  test_value = ({
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'None ,0.00000,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  })
```
cderb: Same question, why only testing for improper solver entry?

tests/test_merge_db_functions.py
cderb: I'd like to have the master list resulting from this merge tested for the new key entry.
If keep_keys is false, I would expect that this would now be the only entry for this key.
There is an opportunity to test keep_keys=True as well. In which case, just this algorithm would be different.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+                                copy_only=False,
+                                keep_keys=False,
+                                target_file=None)
+
+
+def test_get_sqlite_table():
+
+  local_path = '/data/utils/test_files/test_gfx90678.db'
+  cnx_from = sqlite3.connect(local_path)
+  perf_rows, perf_cols = get_sqlite_table(cnx_from, 'perf_db')
```
cderb: This function is imported from analyze_parse_db.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+def test_get_sqlite_row():
+
+  local_path = '/data/utils/test_files/test_gfx90678.db'
+  cnx_from = sqlite3.connect(local_path)
+
+  perf_rows, perf_cols = get_sqlite_table(cnx_from, 'perf_db')
+  for row in perf_rows:
+    perf = dict(zip(perf_cols, row))
+
+    cfg_row, cfg_cols = get_sqlite_row(cnx_from, 'config', perf['config'])
```
cderb: function imported from analyze_parse_db

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,337 @@
...
+      'dilation_w': 1,
+      'dilation_d': 0,
+      'bias': 0,
+      'group_count': 1
+  }
+
+  cnx_from = sqlite3.connect(local_path)
+  cnx_to = sqlite3.connect(final_file)
+
+  res, col = get_sqlite_data(cnx_to, 'config', prune_cfg_dims(cfg))
```
cderb, line 592: function imported from analyze_parse_db

cderb (changes requested):

tests/test_merge_db_functions.py
technicalgrp89: Yes, You are correct. As it is reffered in merge_db.py, to cover all API , added the test case.

tests/test_merge_db_functions.py
technicalgrp89: @alexandraBara , I have modified the test more than 1 sampler. where as other tests are checking the bestSolver input values for zero/none for the time value.

tests/test_merge_db_functions.py
technicalgrp89: In the above sample, trying to pass the space as the input == solver name.

tests/test_merge_db_functions.py
technicalgrp89: trying to pass slover name as None

tests/test_merge_db_functions.py
technicalgrp89: ok. removed.

tests/test_merge_db_functions.py
technicalgrp89: for UT testing purpose, I am supplying only one set of values. We can assert on the basis of result values only.

tests/test_merge_db_functions.py
technicalgrp89: assert added to check the result outcome of masterfile

tests/test_merge_db_functions.py
technicalgrp89: Added the the assert to cehck the final file result file content.

tests/test_merge_db_functions.py
technicalgrp89: corrected

tests/test_merge_db_functions.py
technicalgrp89: Yes, You are correct. As it is reffered in merge_db.py, to cover all API , added the test case.

tests/test_merge_db_functions.py
technicalgrp89: best_solver(vals) - vals is set of fdb_key = alg:solver, time, workspace_size, alg; <repeat for algorithms>;...
best_solver -> finds the fastest solver in the passed (input value->Set).Comparing/finding the fastetst solver among more than one Set is outof scope of this functionality. This function it self traverse through the set and returns the appropriate solver/time taken etc.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+def test_parse_text_fdb_name():
+  master_file = '/data/tests/gfx90a_68.HIP.fdb.txt'
+
+  arch, num_cu, final_file, copy_files = parse_text_fdb_name(master_file)
+  assert (arch == "gfx90a")
+  assert (num_cu == '68')
+  assert (final_file == "/data/tests/gfx90a_68.HIP.fdb.txt")
+  assert (copy_files == [])
```
cderb: Interesting point to be made here.
Our dbs will read ex gfx900_64, splitting arch and num_cu with a '_' up until 64 cus.
Once we go over that number they get mashed into the arch with hex eg, gfx908 120cus = gfx90878
btw gfx90a68 = gfx90a 104cus
parsing here reacted as expected though.
Perhaps we could test both of those cases in this test.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+
+
+def test_parse_text_pdb_name():
+  master_file = '/data/test/gfx1030_36.cd.pdb.txt'
```
cderb: This doesn't look like a typical pdb filename 'gfx1030_36.db' is typical.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvHipImplicitGemmForwardV4R4Xdlops_Padded_Gemm, 0.02352,0,miopenConvolutionFwdAlgoImplicitGEMM,not used',
+      'miopenConvolutionFwdAlgoWinograd':
+          'ConvBinWinogradRxSf2x3g1,0.03856,0,miopenConvolutionFwdAlgoWinograd,not used',
+      'miopenConvolutionFwdAlgoGEMM':
+          'GemmFwdRest,0.05712,2749200,miopenConvolutionFwdAlgoGEMM,not used'
+  })
+
+  solver, time = best_solver(FdbKey_Set)
+  if (len(solver) == 0):
```
cderb: could be: assert len(solver)

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+    assert (False)
+  else:
+    assert (True)
+
+  if (float(time)):
```
cderb: could be: assert float(time)
Is this testing if time != 0 ? That appears to be the effect here.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+        'miopenConvolutionBwdDataAlgoWinograd':
+            'ConvBinWinogradRxSf2x3,0.06365,0,miopenConvolutionBwdDataAlgoWinograd,<unused>',
+        'miopenConvolutionBwdDataAlgoGEMM':
+            'GemmBwdRest,4.0393,32768,rocBlas,<unused>'
+    })
+
+
+def test_best_solver():
+
+  FdbKey_Set = ({
```
cderb: I don't think this name would pass pylint.

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+  assert (time == 0.02352)
+
+  #Sample as space, time 0.0
+  test_value = ({
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          ' ,0.00000,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  })
+
+  new_solver, new_time = best_solver(test_value)
+  if (not (new_solver and new_solver.isspace())):
```
cderb: This condition seems overcomplicated.
If you're expecting that new_solver is a space then: assert new_solver.isspace()
If you're expecting that new_solver is None then: assert not new_solver

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+
+  #Sample as None and time 0.0
+
+  test_value = ({
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          ' ,0.00000,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  })
+
+  new_solver, new_time = best_solver(test_value)
+  if (not (new_solver and new_solver == "None")):
```
cderb: shorten this to an assert

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,477 @@
...
+  key = '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'
+  vals = {
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvMlirIgemmFwdXdlops,0.03776,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  }
+  keep_keys = False
+  target_merge(master_list, key, vals, keep_keys)
+
+  #new key with the avaliable values.
+  if (key == '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'):
```
cderb: Why assert this? This value is set by the test.

cderb (changes requested):

tests/test_merge_db_functions.py
technicalgrp89: yes, as it is set to 0 in test input

tests/test_merge_db_functions.py
technicalgrp89: updated, by removing same.

JehandadKhan:
@technicalgrp89 Please resolve the conflict as well as run the CI

tests/test_merge_db_functions.py
technicalgrp89: added gfx90878 db to test the same

tests/test_merge_db_functions.py
technicalgrp89: added new test case to check the pdb num_cu check.

technicalgrp89:
@cerb review comments updated. Let me know. @alexandraBara

Hildy:
> @cderb review comments updated. Let me know. @alexandraBara

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+  }
+  keep_keys = False
+  target_merge(master_list, key, vals, keep_keys)
+
+  #new key with the avaliable values.
+  if (master_list[key] == {
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvMlirIgemmFwdXdlops,0.03776,0,miopenConvolutionFwdAlgo    ImplicitGEMM,not used'
+  }):
+    assert (True)
```
cderb: Don't do this, this will not get the right result.
```py
if (statement) { assert (True) }
```
 
Do this:
```py
assert statement == True
```

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+          'ConvOclDirectFwd1x1,0.234239,0,miopenConvolutionBwdData    AlgoDirect,not used',
+      'miopenConvolutionBwdDataAlgoGEMM':
+          'GemmBwd1x1_stride1,0.239519,0,miopenConvolutionBwdDataAlgoGEMM,    not used',
+      'miopenConvolutionBwdDataAlgoImplicitGEMM':
+          'ConvAsmImplicitGemmGTCDynamicBwdXdlopsNHWC,0.327199,94633984,miope    nConvolutionBwdDataAlgoImplicitGEMM,not used',
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvMlirIgemmFwdXdlops,0.03776,0,m    iopenConvolutionFwdAlgoImplicitGEMM,not used'
+  }):
+
+    assert (True)
```
cderb: and here

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+
+  key = '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'
+  vals = {}
+  target_merge(master_list, key, vals, keep_keys)
+  assert (vals == {})
+
+  try:
+    target_merge(master_list={}, key={}, vals={}, keep_keys=False)
+  except AttributeError:
+    assert (True)
```
cderb: If you are expecting target_merge to end by raising an AttributeError, you should put an 'assert False' after it, if it doesn't fail, the assert will cause a failure.

as a side-note 'assert True' is not useful, it does nothing

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+
+    vals = {
+        'miopenConvolutionFwdAlgoImplicitGEMM':
+            'ConvMlirIgemmFwdXdlops,0.03776,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+    }
+    master_list = {}
+    target_merge(master_list, key, vals, keep_keys)
+
+  except AttributeError:
+    assert (True)
```
cderb: and here

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+
+  final_file = '/data/tests/final.txt'
+  copy_files = []
+
+  key = '1-48-480-3x3-16-48-480-1-1x1-1x1-1x1-0-NCHW-FP16-F'
+  write_merge_results(master_list, final_file, copy_files)
+
+  if key == open('/data/tests/final.txt').read():
+    assert (True)
+    f.close()
```
cderb: What does this test?
If the key is read from the file:
```py
  assert (True) # a null opp
  f.close()     # where is f defined?

File: tests/test_merge_db_functions.py [hunk 33]
```diff
@@ -0,0 +1,481 @@
...
+  #input set to invalid types
+  master_list = {}
+
+  final_file = '/data/tests/final.txt'
+  key = {0}
+  write_merge_results(master_list, final_file, copy_files)
+
+  if key == open('/data/tests/final.txt').read():
+    assert (True)
+    f.close()
```
cderb: and here

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+  copy_only = False
+  keep_keys = False
+  target_file = '/data/utils/test_files/usr_gfx90a68.HIP.fdb.txt'
+
+  result_file = merge_text_file(master_file, copy_only, keep_keys, target_file)
+
+  if (os.stat(result_file)):
+    assert (True)
+  else:
+    assert (False)
```
cderb: simplify to
```py
assert os.stat(result_file)
```

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+
+  try:
+    assert [row for row in open(master_file)
+           ] == [row for row in open(target_file)]
+  except AssertionError as msg:
+    assert (True)
```
cderb: if master == target then the assert is true and this still passes
correct / simplify to
```py
assert statement == False
```

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+
+  master_file = '/data/utils/test_files/old_gfx90a68.HIP.fdb.txt'
+
+  try:
+    result_file = merge_text_file(master_file,
+                                  copy_only=False,
+                                  keep_keys=False,
+                                  target_file=None)
+  except TypeError:
+    assert (True)
```
cderb: and here

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+def test_get_sqlite_table():
+
+  local_path = '/data/utils/test_files/test_gfx90678.db'
+  cnx_from = sqlite3.connect(local_path)
+  perf_rows, perf_cols = get_sqlite_table(cnx_from, 'perf_db')
+
+  if not perf_rows and not perf_cols:
+    assert (False)
+  else:
+    assert (True)
```
cderb: simplify to one assert

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+    perf = dict(zip(perf_cols, row))
+
+    cfg_row, cfg_cols = get_sqlite_row(cnx_from, 'config', perf['config'])
+    cfg = dict(zip(cfg_cols, cfg_row))
+    cfg.pop('id', None)
+
+  if not cfg_row and cfg_row:
+    assert (False)
+  else:
+    assert (True)
```
cderb: and here

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,481 @@
...
+
+  cnx_from = sqlite3.connect(local_path)
+  cnx_to = sqlite3.connect(final_file)
+
+  res, col = get_sqlite_data(cnx_to, 'config', prune_cfg_dims(cfg))
+
+  if not res:
+    assert (False)
+  else:
+    assert (True)
```
cderb: and here

cderb (changes requested):

tests/test_merge_db_functions.py
technicalgrp89: statement is populated from master file for that perticular key. Just checking the statement would mislead. So, I would store in temp variable and compare.

JehandadKhan:
@technicalgrp89  Can we please drive this PR to conclusion ?

technicalgrp89:
@cderb , @JehandadKhan CI checks are failing, as we need to support following checks in the source merge_db.py
function - target_merge - key set to None-> raise the exception.
              - merge_text_file , target_file set to None -> raise the exception
                   - update_master_list ,loca_paths set to None - > raise the exception
 rest of the review comments are addressed. Let me know.

alexandraBara:
> @cderb , @JehandadKhan CI checks are failing, as we need to support following checks in the source merge_db.py function - target_merge - key set to None-> raise the exception. - merge_text_file , target_file set to None -> raise the exception - update_master_list ,loca_paths set to None - > raise the exception rest of the review comments are addressed. Let mw know.

@technicalgrp89 all of these failures can be checked in their respective functions. 
Please address all 3 by raising a ValueError('some text here') where each failing condition happens. You can check the Tuna code to see an example where we raise a ValueError.
In the utest you can then catch the ValueError for the failing cases and this should complete the code coverage.

If you have Q's about this please reach out.

technicalgrp89:
test_merge_db_functions.py and tuna/utils/merge_db.py are updated for error handling cases. key, master_list, vals and target_file input values are checked and handled for for Null,None,{}."","   " etc. All tests are passing in CI.
@cderb @alexandraBara @JehandadKhan

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,595 @@
...
+
+  vals = {
+      'miopenConvolutionFwdAlgoImplicitGEMM':
+          'ConvMlirIgemmFwdXdlops,0.03776,0,miopenConvolutionFwdAlgoImplicitGEMM,not used'
+  }
+
+  try:
+    target_merge(master_list, key, vals, keep_keys)
+  except AttributeError as e:
+    assert False
```
cderb: Why assert false on attribute error? If the desired outcome is to error on AttributeError, why not just not catch the AttributeError?

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,595 @@
...
+def parse_fdb(filename):
+  find_db = {}
+  fp = open(filename)
+  for line in fp:
+    key, vals = line.split('=')
+    vals = [x.strip() for x in vals.split(';')]
+    find_db[key] = vals
+  fp.close()
+
+  return find_db
```
cderb, line 488: Maybe this belongs in a utility file?

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,595 @@
...
+
+
+def test_get_sqlite_table():
+
+  local_path = "{0}/../utils/test_files/test_gfx90678.db".format(this_path)
+  cnx_from = sqlite3.connect(local_path)
+  perf_rows, perf_cols = get_sqlite_table(cnx_from, 'perf_db')
+
+  if not perf_rows and not perf_cols:
+    assert (False)
```
cderb: equivalent to 
```python
assert perf_rows or perf_cols
```

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,595 @@
...
+  perf_rows, perf_cols = get_sqlite_table(cnx_from, 'perf_db')
+  for row in perf_rows:
+    perf = dict(zip(perf_cols, row))
+
+    cfg_row, cfg_cols = get_sqlite_row(cnx_from, 'config', perf['config'])
+    cfg = dict(zip(cfg_cols, cfg_row))
+    cfg.pop('id', None)
+
+  if not cfg_row and cfg_row:
+    assert (False)
```
cderb: can simplify

tests/test_merge_db_functions.py
```diff
@@ -0,0 +1,595 @@
...
+      'bias': 0,
+      'group_count': 1
+  }
+
+  cnx_to = sqlite3.connect(test_file)
+
+  res, col = get_sqlite_data(cnx_to, 'config', prune_cfg_dims(cfg))
+
+  if not res:
+    assert (False)
```
cderb: can simplify

tuna/utils/merge_db.py
```diff
@@ -207,6 +207,16 @@ def best_solver(vals):
 
 def target_merge(master_list, key, vals, keep_keys):
   """merge for explicit target file"""
+  #Error handling of Ivalid input parameter types
+  if key is None:
+    raise ValueError(f'Invalid key value: {key}')
+  if (not (key and not key.isspace())):
+    raise ValueError(f'Invalid key value: {key}')
+  if (not (key and key.strip())):
+    raise ValueError(f'Invalid key value: {key}')
```
cderb: could simplify to 
```python
if not (key and isinstance(key, str) and key.strip()):
  raise ValueError(f'Invalid key value: {key}')
```
This just asserts that key is a non-empty string. A fairly minimal check.

tuna/utils/merge_db.py
```diff
@@ -207,6 +207,16 @@ def best_solver(vals):
...
+  if not vals:
+    raise ValueError(f'Invalid vals: {vals}')
```
cderb: This check will throw an error when vals is None, False, or an empty set. Not very robust, not sure if useful.

tuna/utils/merge_db.py
```diff
@@ -321,6 +331,12 @@ def write_merge_results(master_list, final_file, copy_files):
 
 def merge_text_file(master_file, copy_only, keep_keys, target_file=None):
   """merge db text files"""
+
+  if target_file is None:
+    raise ValueError(f'Invalid key value: {target_file}')
+  if not target_file and target_file.strip():
+    raise ValueError(f'Invalid key value: {target_file}')
```
cderb: This conditional is problematic.
if not target_file, then target_file is None, False, 0, "", or empty set
if not target_file, then target_file.strip() will be evaluated, which will cause a separate kind of error on every possible value other than "" on which it will have no affect.
Perhaps the objective was to assert that the target_file is a non empty string or perhaps an existing file.

cderb (changes requested):

tests/test_merge_db_functions.py
technicalgrp89: this is a helper function , used in checking the end result of merge_text_file. Originally, defined in test_merge_db.py

tuna/utils/merge_db.py
technicalgrp89: checking target_file , already exist is not part of UT. test targets for valid i/p type

cderb (approved):

alexandraBara (approved):

PR #823, opened 2022-12-02
Automated creating perf_table

alexandraBara:
Added model and framework version to benchmark script
Couple updates on tagging configs
Added function to recreate the conv_perfXYZ table after tuna_gold gets updated.

### DIFF
--- a/tuna/import_benchmark.py
+++ b/tuna/import_benchmark.py
@@ -97,8 +97,14 @@ def parse_args():
                       type=str,
                       dest='file_name',
                       help='File to specify multiple Driver commands')
-  parser.add_argument('--version',
-                      dest='version',
+  parser.add_argument('--md_version',
+                      dest='md_version',
+                      type=str,
+                      default=None,
+                      required=False,
+                      help='Specify model version')
+  parser.add_argument('--fw_version',
+                      dest='fw_version',
                       type=str,
                       default=None,
                       required=False,
@@ -111,14 +117,15 @@ def parse_args():
                       help='Specify model batchsize')
 
   args = parser.parse_args()
-  if args.add_model and not args.version:
+  if args.add_model and not args.md_version:
     parser.error('Version needs to be specified with model')
   if args.add_benchmark and not (args.model and args.framework and
                                  args.gpu_count and args.batchsize and
+                                 args.md_version and args.fw_version and
                                  (args.driver or args.file_name)):
     parser.error(
-        'Model, framework, driver(or filename), batchsize and gpus need to all be'
-        'specified to add a new benchmark')
+        """Model, md_version, framework, fw_version, driver(or filename), batchsize \n
+         and gpus need to all be specified to add a new benchmark""")
   return args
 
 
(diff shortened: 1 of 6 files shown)


tuna/import_benchmark.py
```diff
@@ -111,14 +117,15 @@ def parse_args():
                       help='Specify model batchsize')
 
   args = parser.parse_args()
-  if args.add_model and not args.version:
+  if args.add_model and not args.md_version:
```
iq136boy, line 120: How (where) to check the version info that I can use? I saw in ModelEnum class, it has different model. So I know what model I can use to run the benchmark. Maybe add some comments about the version info in the error message if the version is not support? The same concern for 'fw_version'.

tuna/import_benchmark.py
```diff
@@ -97,8 +97,14 @@ def parse_args():
...
+                      dest='md_version',
+                      type=str,
+                      default=None,
+                      required=False,
+                      help='Specify model version')
+  parser.add_argument('--fw_version',
+                      dest='fw_version',
                       type=str,
                       default=None,
                       required=False,
```
cderb, line 110: Help should be framework version.

tuna/import_benchmark.py
```diff
@@ -158,28 +165,31 @@ def update_frameworks():
...
       except IntegrityError as ierr:
         LOGGER.warning(ierr)
+        session.rollback()
         continue
 
   return True
 
 
-def get_database_id(framework, model, dbt):
+def get_database_id(framework, model, md_version, fw_version, dbt):
```
cderb: Argument order is strange
maybe f, f_v, m, m_v
or f, m, f_v, m_v

tuna/update_golden.py
```diff
@@ -274,6 +311,8 @@ def main():
   total = process_merge_golden(dbt, args.golden_v, query.all())
 
   LOGGER.info("Merged: %s", total)
+  LOGGER.info('Updating conv perf DB table')
+  create_perf_table(args)
```
cderb: This isn't a good place to run this. This script could be invoked more than once while creating a new golden_v.
eg If for a solver tune we use MI100,MI210,MI250 during tuning: We would run this script once to re-base from the previous golden version, then 3 more times to import from each tuning session.
This table would be created 4x and dropped 3x as a result.

cderb (changes requested):

tuna/import_benchmark.py
alexandraBara: hmm, i only have a way to look at frameworks that exist in the DB, but i can duplicate that for the model as well, then you would be able to see whats in the DB for both model and framework.

tuna/update_golden.py
alexandraBara: OK, then ill add an argument to create this table. This way we can specify that argument when needed (aka for the last run of the update) and we would not be creating/deleting multiple times. Would that work?

tuna/import_benchmark.py
alexandraBara: just kidding, this already exists, you can use:
import_benchmark.py --print_models
and it will display name and version

reidkwja (approved):

tuna/update_golden.py
```diff
@@ -274,6 +320,9 @@ def main():
   total = process_merge_golden(dbt, args.golden_v, query.all())
 
   LOGGER.info("Merged: %s", total)
+  LOGGER.info('Updating conv perf DB table')
```
cderb, line 323: Could be inside if.

cderb (approved):

PR #824, opened 2022-12-05
Tweaks for Merge and Export

cderb:
remove expensive queries from export
match merge fdb sorting with export

### DIFF
--- a/tuna/update_golden.py
+++ b/tuna/update_golden.py
@@ -252,7 +252,7 @@ def merge_golden_entries(session, dbt, golden_v, entries, simple_copy=False):
   return count
 
 
-def verif_no_duplicates(entries):
+def verify_no_duplicates(entries):
   """ check entries for duplicates (error in fdb) """
   with DbSession() as session:
     sess_map = sess_info(session)
@@ -370,7 +370,7 @@ def main():
   entries = get_fdb_entries(dbt)
   LOGGER.info("Prepped %s entries", len(entries))
 
-  success = verif_no_duplicates(entries)
+  success = verify_no_duplicates(entries)
   if not success:
     return False
   total = process_merge_golden(dbt, args.golden_v, entries)
--- a/tuna/update_golden.py
+++ b/tuna/update_golden.py
@@ -252,23 +252,20 @@ def merge_golden_entries(session, dbt, golden_v, entries, simple_copy=False):
   return count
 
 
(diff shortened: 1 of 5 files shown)


tuna/update_golden.py
```diff
@@ -109,6 +109,31 @@ def get_fdb_query(dbt):
...
+    attr_str = ','.join(fdb_attr)
+    query = f"select {attr_str} from {dbt.find_db_table.__tablename__}"\
+            f" where valid=1 and session={dbt.session_id};"
+    ret = session.execute(query)
+    entries = []
+    for row in ret:
+      entry = Dummy()
+      for i, col in enumerate(fdb_attr):
+        setattr(entry, col, row[i])
+      entries.append(entry)
```
alexandraBara, line 132: Is this to avoid creating a new SQLAlchemy obj?

cderb: Correct, this replaces the 'get_fdb_query' function and creates a namespace instead of a sqlalchemy orm. I discovered that this made a difference of 2 hours when handling the full gfx90878 dataset (shorter time was 15 min).

tuna/update_golden.py
```diff
@@ -227,14 +252,39 @@ def merge_golden_entries(session, dbt, golden_v, entries, simple_copy=False):
   return count
 
 
+def verif_no_duplicates(golden_v, entries):
```
alexandraBara: since this raises a ValueError I think that should be caught in main followed by a cleaner exit.
This function need a utest, same with get_fdb_entries

alexandraBara (changes requested):

tuna/update_golden.py
cderb: updated

tuna/update_golden.py
```diff
@@ -227,14 +252,36 @@ def merge_golden_entries(session, dbt, golden_v, entries, simple_copy=False):
   return count
 
 
+def verif_no_duplicates(entries):
```
JehandadKhan:
```suggestion
def verify_no_duplicates(entries):
```

JehandadKhan (commented):
@alexandraBara Please review the changes to the various queries and verify them.

JehandadKhan (approved):

JehandadKhan:
@alexandraBara  Please help review this PR

alexandraBara (approved):

PR #826, opened 2022-12-08
SolverAnalytics Grafana Dashboard

msaudulhassan:
1. Create a table we can dump the results from SolverAnalytics in
2. Setup a Grafana dashboard to display that table 

This PR is in parallel to the SolverAnalyics PR https://github.com/ROCmSoftwarePlatform/SolverAnalytics/pull/16


tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,37 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+                                     "filter",
+                                     "padding",
+                                     "stride",
+                                     "layout",
+                                     "precision",
+                                     "direction",
+                                     name="uq_idx"),)
+  
+  # columns definitions
+  golden_v = Column(Integer, nullable=False)
```
alexandraBara: is this the same as the golden_miopen_v? If so, can you please rename it to match the col that tuna uses?

tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,37 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+  filter = Column(String(32), nullable=False)
+  padding = Column(String(32), nullable=False)
+  stride = Column(String(32), nullable=False)
+  dilation = Column(String(32), nullable=False)
```
alexandraBara: im pretty sure these need to be ints. Please check the conv_config table in tuna to see how we define these.

tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,37 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+  layout = Column(String(8), nullable=False)
+  precision = Column(String(8), nullable=False)
```
alexandraBara: also probably not a string. What values does this take? ints? floats?

tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,37 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+  direction = Column(String(1), nullable=False)
```
alexandraBara: are you saving 'F', 'B', 'W'? or 1,2,4? if its the ladder please make it an int.

tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,37 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+  sf = Column(String(length=128), nullable=False) # fastest solver
```
alexandraBara: this needs to be an Fkey into the solver table. Check miopen_tables.py for definition of this in other tables. If you want to store the string (so you dont join with the solver table later), then make it a:
 solver      | varchar(128) not string(128)

alexandraBara (changes requested):

tuna/miopen/miopen_tables.py
msaudulhassan: They are going to be strings like "32x64x64"

tuna/miopen/miopen_tables.py
msaudulhassan: Layout will be like "NCHW" or smtg, and precision will be like "FP16" etc.

tuna/miopen/miopen_tables.py
msaudulhassan: It's the former, F,B,W

tuna/miopen/miopen_tables.py
msaudulhassan: It is. I'll change it

tuna/miopen/miopen_tables.py
msaudulhassan: Resolved: https://docs.sqlalchemy.org/en/14/core/type_basics.html#the-camelcase-datatypes (it automatically translates to VARCHAR, as it does in the solver table)

tuna/miopen/miopen_tables.py
alexandraBara: ok, then precision can stay as string

tuna/miopen/miopen_tables.py
alexandraBara: oh ok, if thats the case for all of them they can stay strings

alexandraBara (approved):

tuna/dbBase/base_class.py
```diff
@@ -71,4 +71,17 @@ def __repr__(self):
     return f"Table name: {self.__table__}\nTable columns: {self.__table__.columns}"
 
 
+class MINIMALBASE():
```
alexandraBara: What is this class? id is already defined in the BASE class. Since this table will reside in the Tuna DB it needs to inherit BASE. Please remove this class.

tuna/miopen/tables.py
```diff
@@ -101,3 +102,5 @@ def set_tables(self):
       self.kernel_cache = ConvolutionKernelCache
       self.golden_table = ConvolutionGolden
       self.benchmark = ConvolutionBenchmark
+
+    self.solver_analytics_results = SolverAnalyticsResults
```
alexandraBara: your table looks at convolutions, please move this on line 105 inside the else

alexandraBara (changes requested):

tuna/dbBase/base_class.py
msaudulhassan: By inheriting base, the table gets the `valid` column, which doesn't make sense for the analytics. So I created this minimal base class with only the `id` column and no additional stuff. But I can remove it (albeit at the expensive of there being redundant/confusing columns in the solver_analytics_results table)?

msaudulhassan: I can subclass `BASE` from `MINIMALBASE` to get rid of redundant code?

msaudulhassan:
The dashboard is at: http://ginger:3000/d/miGbCrF4z/solveranalytics

alexandraBara (changes requested):
Please declare a ConvSolverAnalyticsMixin table that contains all but the 'sa' column. Similar to all our other Mixin tables in Tuna. Declare the 'sa' col, tablenames and UniqueConstraints in the extended tables.

msaudulhassan:
> Please declare a ConvSolverAnalyticsMixin table that contains all but the 'sa' column. Similar to all our other Mixin tables in Tuna. Declare the 'sa' col, tablenames and UniqueConstraints in the extended tables.

Done!

alexandraBara (approved):

tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,78 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+                                     "num_cu",
+                                     "opencl",
+                                     "filter",
+                                     "padding",
+                                     "stride",
+                                     "dilation",
+                                     "layout",
+                                     "precision",
+                                     "direction",
+                                     "sf",
```
cderb, line 610: Assuming there should only be one fastest solver, solver name shouldn't be in the unique list. This would allow there to be multiple entries with separate solver names.

tuna/miopen/miopen_tables.py
```diff
@@ -569,6 +569,78 @@ class BNBenchmark(BASE, BenchmarkMixin):
...
+                                     "sa",
```
cderb, line 635: Should solver name / alternate be in the unique list?

cderb (changes requested):

tuna/miopen/miopen_tables.py
msaudulhassan: There can be multiple fastest solvers. The table is really telling how many times x was the fastest solver for some convolution problem A, how many times y was the fastest solver for A, and so on. A convolution problem here is defined by "filter", "padding", "stride", "dilation", "layout", "precision", "direction". Notice that it is missing stuff like input size, batch size, etc. So x and y can both be the fastest solvers for the same convolution problem, because the underlying input size and batch size are different.

tuna/miopen/miopen_tables.py
msaudulhassan: yes, for the same reason as last comment.

cderb (approved):

alexandraBara:
@msaudulhassan  what is the status of this? Can you please request for this to be merged if this is still the path you are taking with the dashboard? Otherwise, please close PR

PR #829, opened 2023-01-06
tuna/build_driver_cmd.py -> type annotation

technicalgrp89:
build_driver_cmd.py , type hints are added
added mypy cmd execution under utils.groovy->runLint()

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ To create the database run the following script:
 ```
 
 The installation and setup are now complete. To start a tuning cycle, please follow the steps
-documented in [TuningCycle](https://github.com/ROCmSoftwarePlatform/MITuna/blob/develop/doc/TuningCycle.md)
+documented in [TuningCycle](https://github.com/ROCmSoftwarePlatform/MITuna/blob/develop/doc/src/TuningCycle.md)
 
 Code formatting
 ---------------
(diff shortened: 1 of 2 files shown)


vars/utils.groovy
```diff
@@ -541,6 +541,7 @@ def runLint() {
           tuna_docker.inside("") {
             sh "cd tuna && pylint -f parseable --max-args=8 --ignore-imports=no --indent-string='  ' *.py miopen/*.py"
             sh "cd tuna && mypy analyze_parse_db.py"
+            sh "cd tuna && mypy build_driver_cmd.py --ignore-missing-imports"
```
JehandadKhan: Why is `ignore-missing-imports` needed ?

JehandadKhan (changes requested):

vars/utils.groovy
technicalgrp89: To suppress all missing import errors , we need to enable while checking the data types. so that we can ignore dependent/linked libraries. these flags would avoid checking the type annotation of import modules/libs.

technicalgrp89:
@JehandadKhan @alexandraBara @cderb Let me know for the review update

JehandadKhan (approved):

alexandraBara (approved):

PR #832, opened 2023-01-11
New lib example

alexandraBara:
Added new library 'example'. Thus far it can:
./go_fish.py example --add_tables --> creates needed tables for this lib
./go_fish.py example --init session -l somelabel--> add new session to its own 'session_example' table.

Moved code out of MIOpen so it can be reused by example. Lots of functions now call a super function that sets miopen/example common variables.

Added utest: test_example.py
Modified a couple old pytest to accommodate for new code.

New folder: tuna/example --> this contains all example needed classes/scripts. Trying to make this as simple as possible, so if you see code that can be moved into interfaces/helpers let me know. Its important that we make adding a new lib as simple as possible.


tuna/example/example_tables.py
```diff
@@ -0,0 +1,83 @@
...
+  completed = 4
+  timeout = 5
+  error_status = 6
+  no_update = 7
+  errored = 8
+  bad_param = 9
+  error = 10
+  transfer_error = 11
+  not_applicable = 20
+  aborted = 21
```
alexandraBara: @cderb this might be overkill, should be simplify?

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+    parser.add_argument('--init_session',
+                        action='store_true',
+                        dest='init_session',
+                        help='Set up a new tuning session.')
+
+    clean_args('EXAMPLE', 'example')
+    args = parser.parse_args()
+    if len(sys.argv) == 1:
+      parser.print_help()
+      sys.exit(-1)
```
alexandraBara: @cderb im not a fan of this. Can you think of a diff way to go about it? The purpose is to force user to run like:
./go_fish.py miopen/example --some_arguments..

tuna/example/example_tables.py
cderb: You can remove 7 and up, new teams will likely bring their own error statuses. 1-6 should be sufficient as an example.

tuna/example/example_lib.py
cderb: I think this is fair enough. Help message should just ask to supply a library and give options for loaded ones.
btw previously this call hasn't been returning as expected, not sure if it was fixed here.
```sh
./go_fish.py miopen -h
```

alexandraBara: This is fixed, current behavior is:
```
./go_fish.py miopen --help
2023-01-12 08:48:40,530 - go_fish - INFO -  [go_fish.py:60] - ['./go_fish.py', 'miopen', '--help']
usage: go_fish.py [-h] [--add_tables] [{miopen,example}]

positional arguments:
  {miopen,example}  Specify library to run, defaults to MIOpen

optional arguments:
  -h, --help        show this help message and exit
  --add_tables      Add library specific tables. If no library specified, MIOpen is implied
```

tuna/example/example_worker.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class ExampleWorker(WorkerInterface):
+  """ The Example class implementes the worker class. Its purpose is to run a command. It picks up
+  new jobs and when completed, sets the state to compiled. """
+
+  def close_job(self):
+    """mark a job complete"""
+    self.set_job_state('compiled')
+
```
cderb: Don't need to include this one. Only exists as a synchronization measure in fin_class.

tuna/example/example_tables.py
```diff
@@ -0,0 +1,83 @@
...
+
+  reason = Column(String(length=60), nullable=False, server_default="")
+  state = Column(Enum(JobEnum), nullable=False, server_default="new")
+  retries = Column(Integer, nullable=False, server_default="0")
+  result = Column(Text, nullable=True)
+  gpu_id = Column(Integer, nullable=False, server_default="-1")
+  machine_id = Column(Integer, nullable=False, server_default="-1")
+
+
+def get_tables():
```
reidkwja, line 68: Would it be a good idea to add logging to this function to ensure proper tracing of data retrieval process?

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+
+    args_check(args, parser)
+
+    return args
+
+  def launch_worker(self, gpu_idx, f_vals, worker_lst, args):
```
reidkwja: Logging or error messages can help for this function...Usually with launching a session/worker is where things tend to break and perhaps detailed breakdown - with logging - on what failed, it can help with debugging or configuration for the user endpoint

reidkwja (changes requested):

tuna/example/example_worker.py
cderb: Don't need this. close_job is a synchronization function for fin_class, and will just be confusing here.

cderb (approved):

tuna/example/example_tables.py
alexandraBara: i dont think its necessary here. There is logging when the retrieved tables are created

tuna/example/example_lib.py
alexandraBara: the worker class has logging, so all the function calls in this function log info

reidkwja (approved):

tests/test_example.py
```diff
@@ -0,0 +1,62 @@
...
+from tuna.machine import Machine
+from tuna.sql import DbCursor
+from tuna.utils.db_utility import connect_db
+from tuna.example.example_lib import Example
+from tuna.db_engine import ENGINE
+from tuna.utils.logger import setup_logger
+from utils import ExampleArgs
+from tuna.utils.miopen_utility import load_machines
+from tuna.dbBase.sql_alchemy import DbSession
+from tuna.example.session import SessionExample
```
JehandadKhan: Are all these imports used in the example ?

tests/test_fin_builder.py
```diff
@@ -86,30 +86,31 @@ def test_fin_builder():
...
-
-  args.session_id = add_test_session()
+  args.label = 'tuna_pytest_fin_builder'
+  args.session_id = add_test_session(label='tuna_pytest_fin_builder')
+  miopen = MIOpen()
+  miopen.args = args
 
   #update solvers
-  kwargs = get_worker_args(args, machine)
+  kwargs = get_worker_args(args, machine, miopen)
```
JehandadKhan:
```suggestion
  kwargs = miopen.get_worker_args(args, machine)
```

tests/test_fin_builder.py
```diff
@@ -86,30 +86,31 @@ def test_fin_builder():
   args = GoFishArgs()
   machine_lst = load_machines(args)
   machine = machine_lst[0]
-
-  args.session_id = add_test_session()
+  args.label = 'tuna_pytest_fin_builder'
+  args.session_id = add_test_session(label='tuna_pytest_fin_builder')
+  miopen = MIOpen()
+  miopen.args = args
```
JehandadKhan: Is this needed ?

tests/test_worker.py
```diff
@@ -64,15 +64,16 @@ def add_job(w):
   machine = machine_lst[0]
 
   #update solvers
-  kwargs = get_worker_args(args, machine)
+  miopen = MIOpen()
+  miopen.args = args
+  kwargs = get_worker_args(args, machine, miopen)
```
JehandadKhan, line 69: Same as above!

tests/utils.py
```diff
@@ -99,9 +100,26 @@ class GoFishArgs():
   reset_interval = None
   dynamic_solvers_only = False
   label = 'pytest'
-  docker_name = None
-  ticket = None
+  docker_name = 'miopentuna'
```
JehandadKhan, line 103: Check with @cderb if this is still required ?

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+    ])
+
+    parser.add_argument('--init_session',
+                        action='store_true',
+                        dest='init_session',
+                        help='Set up a new tuning session.')
+
+    clean_args('EXAMPLE', 'example')
+    args = parser.parse_args()
+    if len(sys.argv) == 1:
```
JehandadKhan, line 69: Perhaps [this would ](https://stackoverflow.com/questions/7869345/how-to-make-python-argparse-mutually-exclusive-group-arguments-without-prefix)be a better solution

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+        TunaArgs.RESTART_MACHINE, TunaArgs.DOCKER_NAME
+    ])
+
+    parser.add_argument('--init_session',
+                        action='store_true',
+                        dest='init_session',
+                        help='Set up a new tuning session.')
+
+    clean_args('EXAMPLE', 'example')
+    args = parser.parse_args()
```
JehandadKhan:
```suggestion
    self.args = parser.parse_args()
```

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+    if len(sys.argv) == 1:
+      parser.print_help()
+      sys.exit(-1)
+
+    args_check(args, parser)
+
+    return args
```
JehandadKhan:
```suggestion
    return 
```

tuna/example/example_lib.py
JehandadKhan:
```suggestion
  def launch_worker(self, gpu_idx, f_vals, worker_lst):
```

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+    worker = ExampleWorker(**kwargs)
+    if args.init_session:
+      SessionExample().add_new_session(args, worker)
+      return False
+
+    worker.start()
+    worker_lst.append(worker)
+    return True
+
+  def compose_worker_list(self, res, args):
```
JehandadKhan:
```suggestion
  def compose_worker_list(self, better_var_name, args):
```

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+      worker_ids = super().determine_num_procs(machine)
+      if len(worker_ids) == 0:
+        return None
+
+      f_vals = super().get_f_vals(machine, worker_ids, self.args)
+
+      for gpu_idx in worker_ids:
+        self.logger.info('launch mid %u, proc %u', machine.id, gpu_idx)
+        if not self.launch_worker(gpu_idx, f_vals, worker_lst, args):
+          break
```
JehandadKhan:
```suggestion
        if not self.launch_worker(gpu_idx, f_vals, worker_lst, args):
          raise ValueError('The worker failed to launch')
```

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+
+    return worker_lst
+
+  def add_tables(self):
+    ret_t = create_tables(get_tables())
+    self.logger.info('DB creation successful: %s', ret_t)
+
+    return True
```
JehandadKhan, line 125: What is the difference between `add_tables` and `create_tables` ?

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+
+  def run(self):
+    # pylint: disable=duplicate-code
+    """Main function to launch library"""
+    res = None
+    self.args = self.parse_args()
```
JehandadKhan:
```suggestion
    self.parse_args()
```

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+    res = load_machines(self.args)
+    res = self.compose_worker_list(res, self.args)
```
JehandadKhan: If these methods are part of the class, the `self.args` can be removed

tuna/example/example_lib.py
```diff
@@ -0,0 +1,149 @@
...
+    return res
+
+  def get_envmt(self, args):
```
JehandadKhan:
```suggestion
  def get_envmt(self):
```

tuna/example/example_tables.py
JehandadKhan:
```suggestion
  new = 1
  started = 2
  running = 3
  completed = 4
  error = 5
```

JehandadKhan (changes requested):
Stay tuned,  more to come

tuna/example/example_worker.py
```diff
@@ -0,0 +1,71 @@
...
+class ExampleWorker(WorkerInterface):
+  """ The Example class implementes the worker class. Its purpose is to run a command. It picks up
+  new jobs and when completed, sets the state to compiled. """
+
+  def step(self):
+    """Main functionality of the builder class. It picks up jobs in new state and compiles them"""
+
+    if not self.get_job("new", "compile_start", True):
+      sleep(random.randint(1, 10))
```
JehandadKhan, line 44: Add a comment here, explaining why it is necessary

tuna/example/example_worker.py
```diff
@@ -0,0 +1,71 @@
...
+      return False
+
+    self.logger.info('Acquired new job: job_id=%s', self.job.id)
+    self.set_job_state('compiling')
+    cmd_output = self.run_cmd()
+
+    failed_job = True
+    result_str = ''
+    if cmd_output:
+      failed_job = False
```
JehandadKhan, line 54: This implies if the command prints anything, its an error. Would that be generally true?

tuna/example/example_worker.py
```diff
@@ -0,0 +1,71 @@
...
+
+    if failed_job:
+      self.set_job_state('errored', result=result_str)
+    else:
+      self.set_job_state('compiled', result=result_str)
+    return True
+
+  def run_cmd(self):
+    """Run a command"""
```
JehandadKhan:
```suggestion
    """Run the actual workload."""
```

tuna/go_fish.py
```diff
@@ -43,8 +43,13 @@ def parse_args():
                       nargs='?',
                       default=Library.MIOPEN,
                       type=Library,
-                      help="Specify library to run",
+                      help="Specify library to run, defaults to MIOpen",
```
JehandadKhan, line 46: I would suggest remove the default and make it a required parameter

tuna/go_fish.py
```diff
@@ -43,8 +43,13 @@ def parse_args():
...
                       choices=Library)
+  parser.add_argument('--add_tables',
+                      dest='add_tables',
+                      action='store_true',
+                      help='Add library specific tables. If no library specified, '\
+                           'MIOpen is implied')
```
JehandadKhan: Perhaps use `mutually_exclusive_group` from `argparse` to address this.

tuna/miopen/miopen_lib.py
```diff
@@ -57,7 +58,9 @@ def parse_args(self):
     parser = setup_arg_parser(
         'Run Performance Tuning on a certain architecture', [
             TunaArgs.ARCH, TunaArgs.NUM_CU, TunaArgs.VERSION,
-            TunaArgs.CONFIG_TYPE, TunaArgs.SESSION_ID
+            TunaArgs.CONFIG_TYPE, TunaArgs.SESSION_ID, TunaArgs.MACHINES,
+            TunaArgs.REMOTE_MACHINE, TunaArgs.LABEL, TunaArgs.RESTART_MACHINE,
+            TunaArgs.DOCKER_NAME
```
JehandadKhan, line 63: Would it make sense to move the ownership or arguments to MIOpen? Also, is `DOCKER_NAME` used anywhere ?

tuna/miopen/miopen_lib.py
```diff
@@ -170,7 +144,7 @@ def parse_args(self):
                        default=None,
                        help='execute on each machine')
 
-    self.clean_args()
+    clean_args()
```
JehandadKhan, line 145: If `args` is owned by `MIOpen` then `clean_args` should be part of it as well ?

tuna/miopen/miopen_lib.py
```diff
@@ -253,7 +211,7 @@ def do_fin_work(self, args, gpu, f_vals):
       @param gpu Unique ID of the GPU
       @param f_vals Dict containing runtime information
     """
-    kwargs = get_kwargs(gpu, f_vals, args)
+    kwargs = self.get_kwargs(gpu, f_vals, args)
```
JehandadKhan:
```suggestion
    kwargs = self.get_kwargs(gpu, f_vals)
```

tuna/miopen/miopen_lib.py
```diff
@@ -272,7 +230,7 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst, args):
     """
     # pylint: disable=too-many-branches
     worker = None
-    kwargs = get_kwargs(gpu_idx, f_vals, args)
+    kwargs = self.get_kwargs(gpu_idx, f_vals, args)
```
JehandadKhan:
```suggestion
    kwargs = self.get_kwargs(gpu_idx, f_vals)
```

tuna/mituna_interface.py
```diff
@@ -122,3 +124,83 @@ def execute_docker(self, worker, docker_cmd, machine):
     #  self.logger.info(line.rstrip())
 
     return True
+
+  def add_tables(self):
+    """Add library specific tables"""
+    return self.add_tables()
+
+  def determine_num_procs(self, machine):
```
JehandadKhan:
```suggestion
  def get_num_procs(self, machine):
```

tuna/mituna_interface.py
```diff
@@ -122,3 +124,83 @@ def execute_docker(self, worker, docker_cmd, machine):
...
+    """Determine number of processes by compute capacity"""
+    worker_ids = None
+    env = get_env_vars()
+    if env['slurm_cpus'] > 0:
+      num_procs = int(env['slurm_cpus'])
+    else:
+      # JD: This sould be the responsibility of the machine class
+      num_procs = int(machine.get_num_cpus() * .6)
```
JehandadKhan: Either move it to machine, or remove the comment.

tuna/mituna_interface.py
```diff
@@ -122,3 +124,83 @@ def execute_docker(self, worker, docker_cmd, machine):
...
+    f_vals["barred"] = Value('i', 0)
+    f_vals["bar_lock"] = Lock()
+    #multiprocess queue for jobs, shared on machine
+    f_vals["job_queue"] = mpQueue()
+    f_vals["job_queue_lock"] = Lock()
+    f_vals["result_queue"] = mpQueue()
+    f_vals["result_queue_lock"] = Lock()
+    f_vals["machine"] = machine
+    f_vals["b_first"] = True
+    f_vals["end_jobs"] = Value('i', 0)
```
JehandadKhan: As discussed, this could be removed and the objects which have these attributes be passed instead.

tests/test_fin_builder.py
alexandraBara: this is just a helper in tests. This function does not exist in /tuna. There is no point in making this a class function since there are some more steps in between in each miopen/example lib.

JehandadKhan (changes requested):

tuna/example/example_lib.py
alexandraBara: break is actually the intended action here, its only thrown in special cases such as --init_session that only runs once or execute a special cmd etc. Any other times the worker actually fails to launch, the worker will throw its own error in the worker class not the lib class.

tuna/example/example_lib.py
alexandraBara: create tables contains the SQL code to actually create the tables in the DB, this is just a helper function from utils.
Add tables is a function specific to each lib that uses get_tables to get lib specific tables.

tuna/example/example_tables.py
alexandraBara: this is updated in PR #2.

tuna/example/example_worker.py
alexandraBara: This is updated in PR#2 when we can actually execute a cmd.

tuna/go_fish.py
alexandraBara: cant make a positional argument required. Should i remove positional?

tuna/go_fish.py
alexandraBara: Moved to individual lib

tuna/mituna_interface.py
alexandraBara: This is a big change. Can i make this a separate issue and address at a different time? There have already been many parallel updates to source code improvement in general, unrelated to adding a new lib.

tuna/miopen/miopen_lib.py
alexandraBara: its a generic function because its also called by example and pylint complains of duplicate code

tuna/miopen/miopen_lib.py
alexandraBara: yeah, i dont really like how arguments are split between all scripts/miopen/example/go_fish but i dont want to address this here. Ill try and do some improvements in a future PR where i introduce some yaml parsing

tuna/example/example_lib.py
JehandadKhan: Please add the above to their respective doc strings

tuna/go_fish.py
JehandadKhan: The `nargs=?` above makes the positional argument optional, removing it as well as the default to make this argument required. [As described in this SO post](https://stackoverflow.com/questions/4480075/argparse-optional-positional-arguments)

tuna/miopen/miopen_lib.py
JehandadKhan: Lets revisit this when you get to it

JehandadKhan (approved):

tuna/example/example_lib.py
alexandraBara: it wont work, if we force one or the other arg then we cannot print the help

JehandadKhan:
Remaining issues to be addressed in the next PR

PR #833, opened 2023-01-18
Reduced sqlalchemy orm usage + Tuning updates

cderb:
worker_interface and miopen/fin_* updated to use types.SimpleNamespace instead of sqlalchemy objects
These objects have the same attribute sets as the sqlalchemy tables, but otherwise no features. Tracking is added for the object features and insert/update/select functions have been added to utils/db_utility.py.

The use of the 'pending' queues has been suspended, though the logic has been left in for the moment. inserts are now committed as soon as they are ready, by their originating thread.

Work fetching query now untangled from Job query. Config is attached to the Job by fin_class in a subsequent query.

update .jenkins/init_session to optionally take arch / num_cu


tuna/utils/db_utility.py
```diff
@@ -153,6 +158,85 @@ def session_retry(session, callback, actuator, logger=LOGGER):
   return False
 
 
+def gen_update_query(obj, attribs, tablename):
+  """Create a query string updating all attributes for the input object"""
+  set_arr = []
+  for attr in attribs:
+    val = getattr(obj, attr)
+    if val is None:
```
alexandraBara, line 166: is this needed? dont we just default to NULL anyway in the DB col?

tuna/utils/db_utility.py
```diff
@@ -153,6 +158,85 @@ def session_retry(session, callback, actuator, logger=LOGGER):
   return False
 
 
+def gen_update_query(obj, attribs, tablename):
+  """Create a query string updating all attributes for the input object"""
```
alexandraBara: Can you be a little more specific with this description? Is this used for a particular table or all tables that need updating?

tuna/utils/db_utility.py
```diff
@@ -153,6 +158,85 @@ def session_retry(session, callback, actuator, logger=LOGGER):
...
+  val_list = []
+  for attr in attr_list:
+    val = getattr(obj, attr)
+    if val is None:
+      val = 'NULL'
+    elif isinstance(val, str) or isinstance(val, datetime):
+      val = f"'{val}'"
+    elif isinstance(val, bytes):
+      val = val.decode('utf-8')
+      val = f"'{val}'"
```
alexandraBara: this looks like the exact code as the function above. can you merge the some parts or the entire for loop into a separate function and call it from both gen_insert and get_update? Otherwise we have to maintain the same code twice

tuna/utils/db_utility.py
```diff
@@ -153,6 +158,85 @@ def session_retry(session, callback, actuator, logger=LOGGER):
...
+    LOGGER.info('select_row: %s', row)
+    entry = types.SimpleNamespace()
+    for i, col in enumerate(attribs):
+      setattr(entry, col, row[i])
+    entries.append(entry)
+  return entries
+
+
+def has_attr_set(obj, attribs):
+  """test if a namespace as the supplied attributes"""
```
alexandraBara: Is there a typo here in the comment?

tuna/utils/db_utility.py
```diff
@@ -153,6 +158,85 @@ def session_retry(session, callback, actuator, logger=LOGGER):
...
+
+def get_class_by_tablename(tablename):
+  """use tablename to find class"""
+  for c in BASE._decl_class_registry.values():
+    if hasattr(c, '__tablename__') and c.__tablename__ == tablename:
+      return c
+
+
 class DB_Type(enum.Enum):  # pylint: disable=invalid-name ; @chris rename, maybe?
   """@alex defines the types of databases produced in tuning sessions?"""
```
alexandraBara, line 243: Can you please update this comment to:
'''Defines a type for the following MIOpen tables: find_db, kern_db and perf_db"""
Can you also please rename DB_Type to some pylint satisfactory name and remove the disable and comment after.

tuna/worker_interface.py
```diff
@@ -36,14 +36,17 @@
 import socket
 import random
 import string
+import types
 from time import sleep
-from sqlalchemy.exc import IntegrityError, OperationalError  #pylint: disable=wrong-import-order
+from sqlalchemy.exc import IntegrityError, OperationalError, NoInspectionAvailable  #pylint: disable=wrong-import-order
```
alexandraBara: can you please double check if the wrong-import-order is still needed?

tuna/worker_interface.py
```diff
@@ -95,6 +98,15 @@ def __init__(self, **kwargs):
     #initialize tables
     self.set_db_tables()
 
+    self.job = types.SimpleNamespace()
+    try:
+      self.job_attr = [column.name for column in inspect(self.dbt.job_table).c]
+      self.job_attr.remove("insert_ts")
+      self.job_attr.remove("update_ts")
+    except NoInspectionAvailable:
+      #pass here for init_session
```
alexandraBara: I dont like silent failures, please add a LOGGER of some kind in case in the future this triggers for something other than init_session

tuna/worker_interface.py
```diff
@@ -271,8 +276,11 @@ def get_job(self, find_state, set_state, imply_end):
                 else:
                   job.machine_id = self.machine.id
 
+                query = gen_update_query(job, self.job_attr,
+                                         self.dbt.job_table.__tablename__)
+                session.execute(query)
+
               session.commit()
```
alexandraBara, line 282: are we still committing anything here now that there is an execute() call? If not, please remove

alexandraBara (changes requested):

alexandraBara (changes requested):
I forgot to add, can you please move export_db to the miopen folder?

tests/test_merge_db_functions.py
```diff
@@ -114,7 +114,7 @@ def test_load_master_list():
   master_file = "{0}/../utils/test_files/old_gfx90a68.HIP.fdb.txt".format(
       this_path)
 
-  basename = os.path.basename(master_file)
+  os.path.basename(master_file)
```
msaudulhassan: It's not being assigned to anything?

tests/test_merge_db_functions.py
```diff
@@ -379,9 +379,9 @@ def test_target_merge():
 
   try:
     target_merge(master_list, key, vals, keep_keys)
-  except ValueError as err:
-    error_found = False
-  assert (err_found)
+  except ValueError:
+    err_found = True
+  assert err_found
```
msaudulhassan, line 383: Lines 382-384 :: maybe you can do something like:
```
except ValueError as err:
  raise AssertionError(str(err))
```

tests/test_merge_db_functions.py
```diff
@@ -399,7 +399,7 @@ def test_update_master_list():
       }
   }
 
-  key = '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'
+  #key = '1-160-698-5x5-64-79-348-1-1x1-2x2-1x1-0-NCHW-FP32-F'
```
msaudulhassan: might want to remove this comment?

tests/test_merge_db_functions.py
```diff
@@ -459,8 +459,8 @@ def test_write_merge_results():
 
   master_file = "{0}/../utils/test_files/old_gfx90a68.HIP.fdb.txt".format(
       this_path)
-  target_file = "{0}/../utils/test_files/usr_gfx90a68.HIP.fdb.txt".format(
-      this_path)
+  #target_file = "{0}/../utils/test_files/usr_gfx90a68.HIP.fdb.txt".format(
```
msaudulhassan: is this comment needed?

tests/test_update_golden.py
```diff
@@ -37,44 +37,84 @@
...
+    query = session.query(ConvolutionGolden)\
+                    .filter(ConvolutionGolden.golden_miopen_v == args.golden_v)\
+                    .filter(ConvolutionGolden.session == session_id)
     res = query.all()
-  assert len(res) is not None
+    assert len(res) == 1
+    assert res[0].params == 'param'
+
+  success = verify_no_duplicates(entries)
+  assert success
```
msaudulhassan: maybe do just `assert verify_no_duplicates(entries)`

tests/test_update_golden.py
```diff
@@ -37,44 +37,84 @@
...
+  fdb_entry5.params = 'something'
+  entries.append(fdb_entry5)
+  assert verify_no_duplicates(entries) == False
 
   args.create_perf_table = True
   assert create_perf_table(args)
+
+
+def copy(dest, src):
+  dest.__dict__ = src.__dict.copy()
```
msaudulhassan: should it be: `dest.__dict__ = src.__dict__.copy()` ?

tuna/miopen/fin_class.py
```diff
@@ -107,23 +135,39 @@ def set_db_tables(self):
...
-  def base_job_query(self, session):
-    """select tables for job query"""
-    query = session.query(self.dbt.job_table, self.dbt.config_table)
-    return query
-
-  def compose_work_query(self, query):
+  def compose_work_objs(self, session, conds):
     """query for fin command and config"""
+    ret = []
     if self.fin_steps:
```
msaudulhassan, line 141: This can bite back at times. The `if` block would be executed if `self.fin_steps` is `None` , or if it is `0`, or if it is `[]`. I think being specific can save from a potential bug. Like, if you want this block to execute only if self.fin_steps is not None, I'd rather say it so: `if self.fin_steps is not None`, instead of just `if self.fin_steps`

tuna/miopen/fin_eval.py
```diff
@@ -63,14 +63,17 @@ def check_gpu(self):
     """Function to check gpu heartbeat"""
     for _ in range(5):
       if self.machine.chk_gpu_status(self.gpu_id):
-        return
+        return True
       self.logger.warning(
           'Unable to detect GPU: %s, sleeping for %s seconds before retry',
           self.gpu_id, 30)
-      sleep(30)
+      #sleep(30)
```
msaudulhassan: redundant comment?

Dockerfile
```diff
@@ -50,7 +50,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -f -y --all
     libncurses5-dev \
     libnuma-dev \
     libpthread-stubs0-dev \
-    llvm-amdgpu \
```
JehandadKhan, line 53: Lets investigate this?

tuna/miopen/fin_utils.py
```diff
@@ -95,6 +97,19 @@ def get_fin_result(status):
   return success, result_str
 
 
+def to_dict(obj):
```
msaudulhassan: Another way might be to subclass `MySimpleNamespace` from `SimpleNamespace`, and define the `to_dict()` method of  `MySimpleNamespace` to:
```
def to_dict(self,):
  ret = vars(self)  #use vars to return SimpleNamespace as a dict
  for key, val in ret.items():
    if isinstance(val, datetime):
      ret[key] = str(val)
  return ret
```
If there is other logic pertaining to SimpleNamespace scattered here and there, it too can all be moved to inside `MySimpleNamespace`, and thus consolidated in one place this way.

tuna/miopen/fin_builder.py
```diff
@@ -28,20 +28,32 @@
...
+from tuna.utils.db_utility import session_retry
 
 
 class FinBuilder(FinClass):
   """ The Builder class implementes the worker class. Its purpose is to compile jobs. It picks up
   new jobs and when completed, sets the state to compiled. """
 
+  def __init__(self, **kwargs):
+    super().__init__(**kwargs)
+    self.jcache_attr = [
```
JehandadKhan, line 50: Perhaps rename `jcache_attr` to `job_cache`

tuna/utils/db_utility.py
cderb: This is a generic function. It will create an update query on the table with 'tablename' for the supplied 'obj', in which it will set the fields declared in 'attribs'.

tuna/utils/db_utility.py
cderb: The object 'obj', may have some attributes that are None, in which case, when we are generically setting all the attributes in 'attribs', we should convert that to an SQL string.

tuna/utils/db_utility.py
cderb: They are similar. There is likely an indirection that could be used to make this a function.

tuna/utils/db_utility.py
cderb: yea, should be 'has'.

tests/test_merge_db_functions.py
cderb: variable wasn't used.

tests/test_merge_db_functions.py
cderb: From my understanding, this test is to raise an error when target_merge does not raise an error.

tests/test_merge_db_functions.py
cderb: not sure if it had a use as a note for something, but the variable was unused so I commented it out for pylint.

tests/test_merge_db_functions.py
cderb: Also set but unused, so commented for pylint. Not sure if it has a purpose. @technicalgrp89

tests/test_update_golden.py
cderb: Fair enough.

tuna/miopen/fin_class.py
cderb: None, 0, and [] are evaluated as False by python. This one is a test against [], which would be the default value of fin_steps. If the array is populated it would branch into the first case here.

tuna/miopen/fin_eval.py
cderb: I updated this to return false, meaning quite the tuna process, since we can no longer reset the machine in the barrier. I'll just clean up these comments.

tuna/miopen/fin_utils.py
cderb: I had initially used SimpleNamespace in an attempt to not create another custom object. There is virtually no difference between SimpleNamespace and defining an empty Class. But I suppose in this instance Having a custom class that defines this behavior would be useful.

tuna/worker_interface.py
cderb: added logging

tuna/worker_interface.py
cderb: appears to no longer be necessary

tuna/worker_interface.py
cderb: Executes must still be committed.

tests/test_merge_db_functions.py
technicalgrp89: yea,  we need to set it False. Line 382-384 need  , setting False correct.

tests/test_merge_db_functions.py
technicalgrp89: earlier, It was used, now in latest version, It is not being used

tests/test_merge_db_functions.py
technicalgrp89: we can remove.

tests/test_merge_db_functions.py
```diff
@@ -367,7 +367,7 @@ def test_target_merge():
   assert (err_found)
 
   # inputs wth master_list={}, error handling
-  err_found = True
+  err_found = False
```
technicalgrp89, line 369: This needs to be set back to True

tests/test_merge_db_functions.py
technicalgrp89: earlier version, had to use this variable for testing purpose, we can remove as now basename is not being used.

technicalgrp89 (dismissed):

tests/test_merge_db_functions.py
msaudulhassan: @cderb So this line can be deleted?

tests/test_merge_db_functions.py
msaudulhassan: Right, but doesn't this seem like a round-about way of throwing an assertion error? The `err_found` variable can be eliminated:
```
try:
  # something
except ValueError as err:
  assert False, str(err)  # or maybe `raise AssertionError(str(err))`
```
Also, an error message will display

tuna/miopen/fin_utils.py
cderb: Added the empty class with to_dict method as requested.

tests/test_merge_db_functions.py
cderb: It's the reverse.
this:
```python
try:
  # something
except ValueError:
  assert False
```
We could accomplish by just running the 'something' and letting it throw its own error. And not bother with the assert.

We want this to throw and error when 'something' doesn't throw an error.
hence
```python
was_an_error = False
try:
  # something
except ValueError:
  was_an_error = True
assert was_an_error
```

tests/test_merge_db_functions.py
cderb: deleted it

tuna/utils/db_utility.py
cderb: out of scope

tests/test_merge_db_functions.py
msaudulhassan: ah, I see! my bad

alexandraBara (approved):

JehandadKhan (approved):

PR #835, opened 2023-01-24
connection.py/config_type.py- Type Annotations enabled

technicalgrp89:
following files are updated with Type Annotations.
  - config_type.py
  - connection.py


technicalgrp89:
@cderb @alexandraBara @JehandadKhan let me know for code review update.

JehandadKhan (changes requested):
~~Please update the Jenkins file so that these files are added to the mypy list.~~

JehandadKhan (approved):

tuna/connection.py
```diff
@@ -132,7 +140,8 @@ def is_alive(self):
         self.logger.error('Checking isAlive when channel does not exist')
       return not self.out_channel.exit_status_ready()
 
-  def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
+  def exec_command_unparsed(self, cmd: str, timeout: int =int(SSH_TIMEOUT),\
+  abort:Union[Any, None]= None) -> Tuple[Any, Any, None]:
```
cderb: Return type here seems to indicate that the third part of the tuple is a None?
return is std_in, std_out, std_err.

tuna/connection.py
```diff
@@ -153,35 +168,43 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
           self.ssh_connect()
+        i_var: Any
+        o_var: Any
+        e_var: None
         i_var, o_var, e_var = self.ssh.exec_command(cmd, timeout=timeout)
+
       except Exception as exc:
-        self.logger.warning('Machine %s failed to execute command: %s', self.id,
-                            cmd)
+        self.logger.warning('Machine %s failed to execute command: %s', id, cmd)
```
cderb: id isn't declared in this function.

tuna/connection.py
```diff
@@ -153,35 +168,43 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
         sleep(retry_interval)
       else:
         self.out_channel = o_var.channel
         return i_var, o_var, e_var
 
-      if abort is not None and chk_abort_file(self.id, self.logger):
-        self.logger.warning('Machine %s aborted command execution: %s', self.id,
-                            cmd)
+      if abort is not None and chk_abort_file(id, self.logger):
+        self.logger.warning('Machine %s aborted command execution: %s', id, cmd)
```
cderb: id isn't declared in this function.

tuna/connection.py
```diff
@@ -153,35 +168,43 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
         return None, None, None
 
     self.logger.error('cmd_exec retries exhausted, giving up')
+
     return None, None, None
 
-  def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
+
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort:Union[Any, None]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[Any, int, Any]:
```
cderb: return values are (integer, StringIO, StringIO)

tuna/connection.py
```diff
@@ -241,36 +263,43 @@ def ssh_connect(self, abort=None):
...
-                         port=self.port,
-                         timeout=SSH_TIMEOUT,
-                         allow_agent=False)
+        self.ssh.connect(  #type: ignore
+            self.hostname,  #type: ignore
+            username=self.user,  #type: ignore
+            password=self.password,  #type: ignore
+            port=self.port,  #type: ignore
+            timeout=SSH_TIMEOUT,
+            allow_agent=False)
```
cderb: why specify type: ignore?

cderb (changes requested):

tuna/connection.py
technicalgrp89: connection.py: note: In member "ssh_connect" of class "Connection":
connection.py:283: error: "Connection" has no attribute "hostname" , to ignore this check from mypy.

tuna/connection.py
technicalgrp89: error: "Connection" has no attribute "id", to handle this mypy check error  , changed it.

tuna/connection.py
technicalgrp89: error: "Connection" has no attribute "id", to handle this mypy check error , changed it.

tuna/connection.py
technicalgrp89: connection.py:216: return value type (got "Tuple[int, Any, None]",
connection.py:248: return value type (got "Tuple[int, Any, Any]"

to accomodate these , return type defined as , Tuple[int, Any, Any]

tuna/connection.py
technicalgrp89: connection.py:172:  return value type (got "Tuple[int, Any, Any]"
connection.py:195:  return value type (got "Tuple[Any, Any, None]"
connection.py:199:  return value type (got "Tuple[None, None, None]"
connection.py:203: return value type (got "Tuple[None, None, None]" 
to cover all above returns, Have to define as Tuple[Any,Any,Any]

technicalgrp89:
@cderb review comments addressed.

tests/test_merge_db_functions.py
```diff
@@ -36,8 +36,8 @@
 from tuna.utils.merge_db import parse_jobline, parse_text_fdb_name, parse_text_pdb_name
 from tuna.utils.merge_db import best_solver, target_merge
 from tuna.utils.merge_db import update_master_list, write_merge_results
-from tuna.utils.merge_db import merge_text_file
-from tuna.utils.merge_db import get_sqlite_table
+from tuna.utils.merge_db import merge_text_file, merge_sqlite_pdb, merge_sqlite_bin_cache
+from tuna.utils.merge_db import merge_sqlite, get_file_list, get_sqlite_table
```
cderb: Why add these functions, they don't appear to be in this file?

tuna/connection.py
cderb: The connection object has an id attribute, see 55 in allowed_keys. However, it may need to be explicitly set to null for mypy to recognize it.
Also nowhere in this function is a variable of name id initialized, so this can't work.

tuna/connection.py
cderb: see previous comment on id

tuna/connection.py
```diff
@@ -241,36 +263,43 @@ def ssh_connect(self, abort=None):
     self.ssh = paramiko.SSHClient()
     self.ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
     for ssh_idx in range(NUM_SSH_RETRIES):
-      if abort is not None and chk_abort_file(self.id, self.logger):
-        self.logger.warning('Machine %s aborted ssh connection', self.id)
+      if abort is not None and chk_abort_file(id, self.logger):
+        self.logger.warning('Machine %s aborted ssh connection', id)
```
cderb: id not valid here, see previous comment

tuna/connection.py
cderb: Are type annotations helpful if we escape them with ignores?
Also see line 55, these self.vars are declared, but may need an explicit declaration for this type checker to function properly.

tuna/connection.py
```diff
@@ -241,36 +263,43 @@ def ssh_connect(self, abort=None):
...
-        retry_interval = randrange(SSH_TIMEOUT)
+        retry_interval = randrange((int(SSH_TIMEOUT)))
         self.logger.warning(
             'Attempt %s to connect to machine %s (%s p%s) via ssh failed, sleeping for %s seconds',
-            ssh_idx, self.id, self.hostname, self.port, retry_interval)
+            ssh_idx,
+            id,
+            self.hostname,  #type: ignore
+            self.port,  #type: ignore
+            retry_interval)  #type: ignore
```
cderb: let's clean up these ignores.

tests/test_merge_db_functions.py
technicalgrp89: Merging might have led these linking issues. In my local branch , I see this missmatch

tuna/connection.py
alexandraBara: @technicalgrp89  these variables need to be initialized to None in the __init__ so pylint doesnt complain. We do this in every class in /tuna, you can check for reference

tuna/connection.py
technicalgrp89: it is initialised in __init__ funciton. now we can use id, along with self key reference.

tuna/connection.py
technicalgrp89: all local variables are initialized in __init__ , this would enable the self key to recognize class variables. mypy thows error , if class memeber variables are not initialized.

technicalgrp89:
@alexandraBara @cderb let me know for code review updates.

tuna/connection.py
```diff
@@ -43,37 +45,47 @@
...
   #pylint: disable=no-member
-
-  def __init__(self, **kwargs):
+  def __init__(self, **kwargs: dict) -> None:
     #pylint
-    self.logger = None
-    self.local_machine = False
-    self.subp = None  # Holds the subprocess obj for local_machine
-    self.out_channel = None  # Holds the out channel for remote connection
+    self.logger: Any = None
```
cderb: I think this is a logging.Logger

tuna/connection.py
```diff
@@ -43,37 +45,47 @@
...
+    self.local_machine: bool = False
+    self.subp: Any = None  # Holds the subprocess obj for local_machine
```
cderb: subprocess.Popen

tuna/connection.py
```diff
@@ -43,37 +45,47 @@
...
+    self.out_channel: Any = None  # Holds the out channel for remote connection
```
cderb: paramiko.channel.Channel

tuna/connection.py
```diff
@@ -43,37 +45,47 @@
...
+
+    #initialize the class member variables
+    self.id = 0  # pylint: disable=invalid-name
+    self.hostname = ' '
+    self.port = 0
+    self.user = None
+    self.password = None
+
+    self.ssh: Any = None
```
cderb: paramiko.SSHClient

tuna/connection.py
```diff
@@ -43,37 +45,47 @@
...
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
 
-    self.ssh = None
-
     if self.logger is None:
       self.logger = setup_logger('Connection')
 
     self.inst_bins = {'which': True, 'cd': True}
-    self.connect(self.chk_abort_file)
+    self.connect(chk_abort_file)
```
cderb: self shouldn't be removed here

tuna/connection.py
```diff
@@ -106,22 +119,25 @@ def get_bin_str(cmd):
 
     return bin_str
 
-  def test_cmd_str(self, cmd):
+  def test_cmd_str(self, cmd: str) -> bool:
     """Function to look for installed binary"""
+    split_cmd: Any
```
cderb: str initially and then list
if we're removing ambiguous typing, change this to two variables.
The first is needed to make a mutable copy of cmd.
The second is for a component list.

tuna/connection.py
```diff
@@ -132,18 +148,25 @@ def is_alive(self):
         self.logger.error('Checking isAlive when channel does not exist')
       return not self.out_channel.exit_status_ready()
 
-  def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
+  def exec_command_unparsed(self, cmd: str, timeout: int =int(SSH_TIMEOUT),\
+  abort:Union[Any, None]= None) -> Tuple[Any, Any, Any]:
```
cderb: @alexandraBara can you comment on typing for abort? It looks like we are using it as a bool, but I haven't found it set at any point.

tuna/connection.py
```diff
@@ -153,35 +176,45 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
 
+    cmd_idx: int
     for cmd_idx in range(NUM_CMD_RETRIES):
       try:
         if self.ssh is None or not self.ssh.get_transport(
         ) or not self.ssh.get_transport().is_active():
           self.ssh_connect()
+        i_var: Any
+        o_var: Any
+        e_var: None
```
cderb: All file-like objects

tuna/connection.py
```diff
@@ -132,18 +148,25 @@ def is_alive(self):
...
     # pylint: disable=broad-except
     if not self.test_cmd_str(cmd):
-      raise ValueError(f'Machine {self.id} failed, missing binary: {cmd}')
+      raise ValueError('Machine {self.id} failed, missing binary: {cmd}')
 
     if self.local_machine:
       #universal_newlines corrects output format to utf-8
       # pylint: disable=consider-using-with ; see exec_command_unparsed's docstring
+      stdout: Any
+      stderr: Any
```
cderb: file-like objects

tuna/connection.py
cderb: Returns are file-like objects, first may be None because we don't capture stdin for Popen.

tuna/connection.py
```diff
@@ -153,35 +176,45 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
-        retry_interval = randrange(SSH_TIMEOUT)
+        retry_interval = randrange((int)(SSH_TIMEOUT))
         self.logger.warning('sleeping for %s seconds', retry_interval)
         sleep(retry_interval)
       else:
         self.out_channel = o_var.channel
         return i_var, o_var, e_var
 
-      if abort is not None and chk_abort_file(self.id, self.logger):
+      if abort is not None and chk_abort_file(id, self.logger):
```
cderb: Note: abort is used here to determine when to call chk_abort_file. I believe this is meant to be self.chk_abort_file, since it is initialized in the constructor through the keys.

cderb: Also this should be self.id not id.

tuna/connection.py
```diff
@@ -153,35 +176,45 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
     return None, None, None
 
-  def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
+
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort:Union[Any, None]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[int, Any, Any]:
     # pylint: disable=too-many-nested-blocks, too-many-branches
     """Function to exec commands"""
+    o_var: Any
+    e_var: None
```
cderb: file-like objects

tuna/connection.py
```diff
@@ -190,23 +223,23 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
       if not proc_line:
         # pylint: disable-next=unnecessary-lambda-assignment ; more readable
         proc_line = lambda x: self.logger.info(line.strip())
-      ret_out = StringIO()
-      ret_err = e_var
+      ret_out: Any = StringIO()
+      ret_err: Any = e_var
```
cderb: file-like objects

tuna/connection.py
```diff
@@ -153,35 +176,45 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
         return None, None, None
 
     self.logger.error('cmd_exec retries exhausted, giving up')
+
     return None, None, None
 
-  def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
+
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort:Union[Any, None]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[int, Any, Any]:
```
cderb: proc_line is a function, if it isn't initialized we define it as a lambda in the body. 
It may be ok to remove it as an input parameter.

cderb: return is int and 2 file-like objects

tuna/connection.py
```diff
@@ -43,37 +45,47 @@
...
 
-  def check_binary(self, bin_str):
+  def check_binary(self, bin_str: str) -> bool:
     """Checking existence of binary"""
     if bin_str in self.inst_bins:
       return self.inst_bins[bin_str]
 
-    cmd = f"which {bin_str}"
+    cmd: str = f"which {bin_str}"
+    out: Any
```
cderb: file-like object

tuna/connection.py
```diff
@@ -153,35 +176,45 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
+        e_var: None
         i_var, o_var, e_var = self.ssh.exec_command(cmd, timeout=timeout)
+
       except Exception as exc:
         self.logger.warning('Machine %s failed to execute command: %s', self.id,
                             cmd)
         self.logger.warning('Exception occurred %s', exc)
         self.logger.warning('Retrying ... %s', cmd_idx)
-        retry_interval = randrange(SSH_TIMEOUT)
+        retry_interval = randrange((int)(SSH_TIMEOUT))
```
cderb: Was there an issue where SSH_TIMEOUT needed to be cast to an int everywhere?
Perhaps change 60.0 to 60 at the top of the file?

cderb (changes requested):

tuna/connection.py
alexandraBara: It looks like a bool correct, i also dont see it set differently anywhere.

alexandraBara:
@technicalgrp89  please address @cderb comments and keep in mind that using type Any defies the purpose of type annotations.

tuna/connection.py
technicalgrp89: Done, it is of type self.logger: Logger

tuna/connection.py
technicalgrp89: Changed to TextIO type

tuna/connection.py
technicalgrp89: correted

tuna/connection.py
technicalgrp89: Changed to TextIO types

tuna/connection.py
technicalgrp89: Changed to TextIO

tuna/connection.py
technicalgrp89: changed id to self.id

tuna/connection.py
technicalgrp89: changed to TextIO-> out

tuna/connection.py
technicalgrp89: added self to check abort call

technicalgrp89: self.chk_abort_file = chk_abort_file added as object attribute set. In init routine

tuna/connection.py
```diff
@@ -106,22 +125,25 @@ def get_bin_str(cmd):
...
+      bin_str: str = self.get_bin_str(sub_cmd)
+      installed: bool = self.check_binary(bin_str)
       if not installed:
         self.logger.error('Tuna cannot find binary: %s', bin_str)
         return False
 
     return True
 
-  def is_alive(self):
+  def is_alive(self) -> Optional[bool]:
```
cderb: Can we set the return type to bool?

tuna/connection.py
```diff
@@ -132,7 +154,56 @@ def is_alive(self):
...
 
-  def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
+  def is_connected(self) -> bool:
+    """ Checks the status of connection """
+    status: bool = True
+    status = self.ssh.get_transport() is not None and \
+    self.ssh.get_transport().is_active() #type: ignore
+    return status
+
+  def connect(self, abort: Any = None) -> None:
```
cderb: abort is a function, not sure if mypy has a type for that.

tuna/connection.py
```diff
@@ -169,19 +249,23 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
         sleep(retry_interval)
       else:
         self.out_channel = o_var.channel
-        return i_var, o_var, e_var
+        return i_var, o_var, e_var  #type: ignore
```
cderb: Why ignore here?

tuna/connection.py
```diff
@@ -169,19 +249,23 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
 
       if abort is not None and chk_abort_file(self.id, self.logger):
         self.logger.warning('Machine %s aborted command execution: %s', self.id,
                             cmd)
-        return None, None, None
+        return None, None, None  #type: ignore
 
     self.logger.error('cmd_exec retries exhausted, giving up')
-    return None, None, None
+    return None, None, None  #type: ignore
```
cderb: Maybe an optional in the return type definition would account for these None values.

tuna/connection.py
```diff
@@ -190,87 +274,49 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
       if not proc_line:
         # pylint: disable-next=unnecessary-lambda-assignment ; more readable
         proc_line = lambda x: self.logger.info(line.strip())
-      ret_out = StringIO()
-      ret_err = e_var
+      ret_out: Optional[TextIO] = StringIO()
+      ret_err: ChannelStderrFile = e_var
+      ret_code: Optional[int] = 0
```
cderb: Non-Optional, is an int

alexandraBara (approved):

tuna/connection.py
```diff
@@ -190,87 +274,49 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
...
           if line == '' and not self.is_alive():
             break
           if line:
             proc_line(line)
-            ret_out.write(line)
-        ret_out.seek(0)
+            if ret_out is not None:
+              ret_out.write(line)
+        if ret_out is not None:
+          ret_out.seek(0)
```
cderb: ret_out is never None, no need for these if statements.

tuna/connection.py
```diff
@@ -190,87 +274,49 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
...
         if self.local_machine:
-          ret_code = self.subp.returncode
-          if ret_out:
+          if self.subp is not None:
+            ret_code = self.subp.returncode
+          else:
+            ret_code = 0
+          if ret_out is not None:
```
cderb: ret_out can't be None

tuna/connection.py
```diff
@@ -190,87 +274,49 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
       if not proc_line:
         # pylint: disable-next=unnecessary-lambda-assignment ; more readable
         proc_line = lambda x: self.logger.info(line.strip())
-      ret_out = StringIO()
-      ret_err = e_var
+      ret_out: Optional[TextIO] = StringIO()
+      ret_err: ChannelStderrFile = e_var
```
cderb: This type would be a union with StringIO, though ideally they would both be converted to StringIO.

cderb (changes requested):

tuna/connection.py
technicalgrp89: But, In retrun we have None, iin position of int. this leads a error in mypy check, int position inretrun types send None too

tuna/connection.py
technicalgrp89: I tried this option, it iwould not resolve

tuna/connection.py
cderb: What line is ret_code set to None?

technicalgrp89: corrected

tuna/connection.py
```diff
@@ -174,14 +257,17 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
       if abort is not None and chk_abort_file(self.id, self.logger):
         self.logger.warning('Machine %s aborted command execution: %s', self.id,
                             cmd)
-        return None, None, None
+        return i_var, o_var, e_var
 
     self.logger.error('cmd_exec retries exhausted, giving up')
-    return None, None, None
+    return i_var, o_var, e_var
```
cderb, line 263: 260, 263 I don't see a path where these values are initialized before these return statements. How is this not a pylint error?

tuna/connection.py
```diff
@@ -190,87 +276,46 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
...
           if line:
             proc_line(line)
             ret_out.write(line)
         ret_out.seek(0)
         if self.local_machine:
-          ret_code = self.subp.returncode
+          if self.subp is not None:
+            ret_code = self.subp.returncode
+          else:
+            ret_code = 0
```
cderb: self.subp is set in exec_command_unparse on the condition that self.local_machine is true. This if / else is unnecessary. please remove.

tuna/connection.py
```diff
@@ -190,87 +276,46 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
...
           if ret_out:
             ret_out.seek(0)
-            ret_err = StringIO()
-            err_str = ret_out.readlines()
+            ret_err = StringIO()  #type: ignore
```
cderb: We could use another union on this variable with StringIO, then we don't need to ignore it. Eventually these should be converted to pure StringIO types.

tuna/connection.py
```diff
@@ -190,87 +276,46 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
...
+            err_str: list = ret_out.readlines()
             for line in err_str[-5:]:
               ret_err.write(line)
             ret_err.seek(0)
             ret_out.seek(0)
         else:
-          ret_code = self.out_channel.recv_exit_status()
+          if self.out_channel is not None and ret_code is not None:
```
cderb: Remove this if, condition is already handled by the else.

tuna/connection.py
```diff
@@ -190,87 +276,46 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
...
+            ret_code = self.out_channel.recv_exit_status()
+
       except (socket.timeout, socket.error) as exc:
         self.logger.warning('Exception occurred %s', exc)
         ret_code = 1
-        ret_out = None
```
cderb, line 219: This changes behavior. Might be ok, but I want to know how it affects the methods that use this function. May need to be an update to read the ret_code first.

tuna/connection.py
```diff
@@ -132,7 +155,57 @@ def is_alive(self):
...
+              self.port, retry_interval)
+          sleep(retry_interval)
+        else:
+          self.logger.info(
+              'SSH connection successfully established to machine %s', self.id)
+          return True
+
+      self.logger.error('SSH retries exhausted machine: %s', self.hostname)
+      return False
+    return False
```
cderb, line 204: I'm assuming this block was just moved and the only changes were to add return types and a couple variable types.

cderb (changes requested):

tuna/connection.py
technicalgrp89: pylint checks are passing.

tuna/connection.py
technicalgrp89: earlier, as It was set to NONE (self.subp), it was raiging mypy NONE type , no attribute error.

tuna/connection.py
technicalgrp89: redundant code is moved to a helper routin is_connected(), rest remains same.

tuna/connection.py
cderb: These values need to be explicitly initialized before they are used as returns. (set to None)

cderb: If you can verify valid behavior from the users of this function you could also set it as an empty IO / Channel object.

tuna/connection.py
```diff
@@ -174,14 +257,18 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
+        return i_var, o_var, e_var
 
     self.logger.error('cmd_exec retries exhausted, giving up')
-    return None, None, None
+    return i_var, o_var, e_var
 
-  def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort: Optional[bool]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[Union[int, Any], Union[TextIO, StringIO, \
+  ChannelFile],Union[ChannelStderrFile, StringIO]]:
```
cderb: return type seems overcomplicated.
tuple val 1 : should always be int
tuple val 2 : appears to be exclusively StringIO
tuple val 3 : appears to be exclusively StringIO

tuna/connection.py
```diff
@@ -190,11 +277,12 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
       if not proc_line:
         # pylint: disable-next=unnecessary-lambda-assignment ; more readable
         proc_line = lambda x: self.logger.info(line.strip())
-      ret_out = StringIO()
-      ret_err = e_var
+      ret_out: Union[TextIO, StringIO] = StringIO()
```
cderb, line 280: StringIO type should be sufficient.

tuna/connection.py
```diff
@@ -174,14 +257,18 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort: Optional[bool]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[Union[int, Any], Union[TextIO, StringIO, \
+  ChannelFile],Union[ChannelStderrFile, StringIO]]:
     # pylint: disable=too-many-nested-blocks, too-many-branches
     """Function to exec commands"""
+    o_var: ChannelFile
+    e_var: ChannelStderrFile
     _, o_var, e_var = self.exec_command_unparsed(cmd, timeout, abort)
     try:
       if not o_var:
```
cderb: Following return type will be a result of POpen in exec_command_unparsed, which would be StringIO. (line 188, 275)

cderb (changes requested):

tuna/connection.py
```diff
@@ -190,11 +277,12 @@ def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
       if not proc_line:
         # pylint: disable-next=unnecessary-lambda-assignment ; more readable
         proc_line = lambda x: self.logger.info(line.strip())
-      ret_out = StringIO()
-      ret_err = e_var
+      ret_out: Union[TextIO, StringIO] = StringIO()
+      ret_err: Union[ChannelStderrFile, StringIO] = e_var
```
cderb: StringIO type should be sufficient.

technicalgrp89: connection.py:281: error: Incompatible types in assignment (expression has type "ChannelStderrFile", variable has type "StringIO")
Found 1 error in 1 file (checked 1 source file)
Either we have to use #type :ignore or add ChannelStderrFile as possible datatype. because, e_var is ChannelStderrFile type.

tuna/connection.py
technicalgrp89: It is returning both StringIO and Channel file objects(input/output/error types) so the return types have to have same type hints. 
Line number : 232 retruns IO objects.
Line number 244: retruns ChannelObjects

technicalgrp89: 254 , out channel is paramiko channel type and ovar.channel doesnot belog to StringIO object type so changing o_var to StringIO doesnot yelied compatible type, line 254 assignment throws error.

connection.py: note: In member "exec_command" of class "Connection":
connection.py:271: error: Incompatible types in assignment (expression has type "ChannelStderrFile", variable has type "StringIO")

tuna/connection.py
cderb: Looks to me like ret_err is set to a StringIO in the try. Looks like this e_var value is only used in case of an exception, so I think it may not be necessary to keep it.

tuna/connection.py
cderb: Disregarding the implications of the type hinting, for reference, this is the real type of the returned values at this point.

tuna/connection.py
technicalgrp89: 269 : o_var is defined as Channel_File , it is need in Tuple value 2
from Line numbers 276 - 308, we start processing lines. In this block, ret_out is defined as union of[textIO, StringIO] = StringIO()

tuna/connection.py
technicalgrp89: removed the assignment as it has no that much importance in this context.

tuna/connection.py
technicalgrp89: connection.py:274: error: Incompatible return value type (got "Tuple[int, ChannelFile, ChannelStderrFile]", expected "Tuple[int, Union[ChannelFile, StringIO], StringIO]")
connection.py:308: error: Incompatible return value type (got "Tuple[Union[int, Any], StringIO, Union[ChannelStderrFile, StringIO]]", expected "Tuple[int, Union[ChannelFile, StringIO], StringIO]")
we need to handle these types. when we combine. retrun statement passes all the tests too. as exec parse return/makes assighment to different IO types, mypy does not support casting.

tuna/connection.py
technicalgrp89: as , it is error exception code handling. I think seeking the file point to zero is fine. Any other part of the code base, just error message is logged  no action taken on any return parameters.

tuna/connection.py
technicalgrp89: In code base, we just log the error messages. and in most of the case we look for mid paramenter that is output file type. some cases , we are checking for exit_recev_status/0ready(), which we are calling already. no where we it is set to empty object type. Eiter truncate/seek(0) can be used.

tuna/connection.py
technicalgrp89: Only in Tuple - 2 Text IO removed.

tuna/connection.py
```diff
@@ -174,14 +257,18 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort: Optional[bool]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[ Union[int, Any], Union[ChannelFile, StringIO], \
+  Union[ChannelStderrFile, StringIO]]:
     # pylint: disable=too-many-nested-blocks, too-many-branches
     """Function to exec commands"""
+    o_var: ChannelFile
+    e_var: ChannelStderrFile
     _, o_var, e_var = self.exec_command_unparsed(cmd, timeout, abort)
     try:
       if not o_var:
```
cderb: Actually this check is a result of the 'None' returns expected from exec_command_unparsed, those were replaced on 260 and 263, this 'if' and 'return' (274 & 275) may be removed.

tuna/connection.py
```diff
@@ -174,14 +257,18 @@ def exec_command_unparsed(self, cmd, timeout=SSH_TIMEOUT, abort=None):
...
-        return None, None, None
+        return i_var, o_var, e_var
 
     self.logger.error('cmd_exec retries exhausted, giving up')
-    return None, None, None
+    return i_var, o_var, e_var
 
-  def exec_command(self, cmd, timeout=SSH_TIMEOUT, abort=None, proc_line=None):
+  def exec_command(self, cmd: str, timeout: int=int(SSH_TIMEOUT), abort: Optional[bool]=None,\
+  proc_line: Union[Any, None]= None) -> Tuple[ Union[int, Any], Union[ChannelFile, StringIO], \
```
cderb: By removing the return at line 275, we can assure the return for this function will be (int, StringIO, StringIO).

cderb (changes requested):

cderb (approved):

alexandraBara (approved):

PR #843, opened 2023-02-03
Enable YAML file

alexandraBara:
Added yaml config file, sample run:

./go_fish.py miopen --yaml miopen/sample.yaml
./go_fish.py example --yaml example/sample.yaml

Note: not all MIOpen functionalities are yet enabled to run through go_fish:
- import_configs
- load_job
- export_db
- import_db
- update_golden

Once these get enabled with a go_fish cmd, they can also be enabled in the yaml file.

For a composite yaml file, such as miopen/sample.yaml and example/sample.yaml, Tuna parses the yaml files and splits it by individual yaml files saved in /tmp for the run.


tuna/go_fish.py
```diff
@@ -40,29 +41,56 @@ def parse_args():
...
   return vars(args)
 
 
 def main():
   """Main function to start Tuna"""
   LOGGER.info(sys.argv)
+  LOGGER.info(len(sys.argv))
   args = parse_args()
-  library = get_library(args)
+  if '--yaml' in sys.argv and len(sys.argv) > 4:
```
cderb: Input error checking should be a part of parse_args()

tuna/yaml_parser.py
```diff
@@ -0,0 +1,107 @@
...
+  """Split initial YAML file to multiple files and return list of files"""
+  yaml_files = []
+  common_yaml_part = get_common_yaml(yaml_dict)
+
+  for key in yaml_dict:
+    #only look at enabled items
+    if key in tuning_steps and yaml_dict.get(key).get('enabled'):
+      #append common args
+      new_yaml = common_yaml_part.copy()
+      yaml_dict[key].pop('enabled', None)
```
cderb, line 98: mutates an input object

tuna/yaml_parser.py
```diff
@@ -0,0 +1,107 @@
...
+
+      #append nested dicts
+      for key2, value2 in yaml_dict.get(key).items():
+        new_yaml[key2] = value2
+
+      if key in single_op:
+        new_yaml[key] = True
+      yaml_files.append(dump_yaml(new_yaml))
```
cderb, line 106: At these point they are yaml dicts?

tuna/go_fish.py
```diff
@@ -40,29 +41,56 @@ def parse_args():
...
+    return False
+
+  #case no yaml file
+  libraries = []
+  libraries.append(get_library(args))
+  yaml_files = [args['yaml']]
+
+  #case with yaml file
+  if args['yaml']:
+    yaml_files = parse_yaml(args['yaml'], args['lib'])
```
cderb, line 75: now yaml_dicts?

tuna/load_job.py
```diff
@@ -110,7 +110,6 @@ def parse_args():
       '--session_id',
       action='store',
       required=True,
-      type=int,
```
cderb, line 113: Why remove int?

tuna/go_fish.py
```diff
@@ -40,29 +41,56 @@ def parse_args():
...
-    if worker_lst is None:
-      return True
-
-    for worker in worker_lst:
-      worker.join()
-      LOGGER.warning('Process finished')
+    for library, yaml_file in zip(libraries, yaml_files):
+      args['yaml_file'] = yaml_file
+      if args['yaml_file']:
+        sys.argv[3] = yaml_file
```
cderb, line 81: Looks like original behavior is preserved when yamle_file is None.

tuna/go_fish.py
```diff
@@ -40,29 +41,56 @@ def parse_args():
...
+  libraries = []
+  libraries.append(get_library(args))
+  yaml_files = [args['yaml']]
+
+  #case with yaml file
+  if args['yaml']:
+    yaml_files = parse_yaml(args['yaml'], args['lib'])
+    libraries = []
+    for _ in range(len(yaml_files)):
+      libraries.append(get_library(args))
```
cderb: Maybe the list of libraries isn't necessary. It seems that they will all be the same since there is only 1 lib argument.

tuna/go_fish.py
```diff
@@ -40,29 +41,56 @@ def parse_args():
...
   try:
-    #returns a list of workers/processes it started
-    worker_lst = library.run()
-    if worker_lst is None:
-      return True
-
-    for worker in worker_lst:
-      worker.join()
-      LOGGER.warning('Process finished')
+    for library, yaml_file in zip(libraries, yaml_files):
```
cderb: Not all of the tuning steps present in the proposed yaml are launched by go_fish. Some are part of independent scripts that do not use a worker_interface. This assumes all yamls will use library.run which composes a worker list.

cderb (changes requested):

tuna/go_fish.py
alexandraBara: hmm like making it an exclusive group? ill investigate

tuna/yaml_parser.py
alexandraBara: yes, the file gets split up in multiple dicts and then they get written to disk as separate yaml files, so basically splitting a yaml file with multiple tuning instructions into separate files with each 1 tuning instruction

tuna/go_fish.py
alexandraBara: here they are files, yaml_files here are actual paths to /tmp/<random_name>.yaml

tuna/load_job.py
alexandraBara: cause u cant have action=store and define a type with jsonargparse ArgParser.

tuna/go_fish.py
alexandraBara: if the yaml files gets split and it turns out there is only 1 tuning instruction then length of libraries is 1. If the initial yaml file has multiple tuning instructions and it gets split into 3 yaml files, then size of libraries will be 3, and each lib will get 1 of the 3 yaml files

tuna/go_fish.py
alexandraBara: yes, i will remove those from the yaml file until they get integrated

tuna/go_fish.py
alexandraBara: correct, is there something here that needs changing?

tuna/yaml_parser.py
alexandraBara: will make a local copy

tuna/go_fish.py
cderb: just confirming.

tuna/go_fish.py
cderb: I mean the input library to go_fish can only ever be 1 library, so why make a list here? Couldn't we just use the 1 library and iterate over the yaml_files? I don't see why we're zipping a list of duplicates with the yaml_files.

tuna/load_job.py
cderb: Delete store and keep int like it is in update_golden?

reidkwja (commented):
Would it be possible to look at an integration with docker's new dev environment tool? I am trying to think of ways where we can emulate Tuna for a local dev without having to SSH all the time. 
https://docs.docker.com/desktop/dev-environments/

tuna/load_job.py
alexandraBara: yeah, ill remove action and keep the type

tuna/go_fish.py
alexandraBara: yes, ill update

cderb (approved):

JehandadKhan (approved):

PR #848, opened 2023-02-09
Tuna Folder Re-Organization

cderb:
Moved miopen only python files out of the base tuna directory.
Organized tuna/miopen folder into new subdirectories:  db, driver, modules, scripts, worker
Updated imports for all files

Temporarily disabling SolverAnalytics, as the directories should be updated with the new paths.


alexandraBara (changes requested):
config_type.py can be moved to miopen as well don't you think?

cderb:
> config_type.py can be moved to miopen as well don't you think?

Wasn't sure if it was a potentially reusable concept. But I'll move it over.

alexandraBara (approved):

Jenkinsfile
```diff
@@ -127,7 +127,8 @@ pipeline {
         agent{  label "tunatest" }
         steps {
           script {
-            utils.solverAnalyticsTest()
+            //utils.solverAnalyticsTest()
+            echo "Need to update paths in SolverAnalytics repo"
```
msaudulhassan: Let me update SolverAnalytics to work with this new folder structure and then let's enable these tests.

msaudulhassan: In fact, the only imports I have from tuna are:
```
from tuna.miopen.tables import MIOpenDBTables
from tuna.config_type import ConfigType
from tuna.utils.db_utility import get_id_solvers
```
Let me know what these would change to, and SA should be able to work with the new tuna structure

msaudulhassan:
I find it odd that the scripts that we actually use (like load_job.py, update_golden.py etc.) are tucked away inside `MITuna/tuna/miopen/subcmd/`, while the utilities are more accessible/visible (in MITuna/tuna/). Would it be a good idea to move the files in MITuna/tuna/ to a subdirectory, and move the scripts inside MITuna/tuna/miopen/subcmd/ to MITuna/tuna/ ?

msaudulhassan (changes requested):

cderb:
> I find it odd that the scripts that we actually use (like load_job.py, update_golden.py etc.) are tucked away inside `MITuna/tuna/miopen/subcmd/`, while the utilities are more accessible/visible (in MITuna/tuna/). Would it be a good idea to move the files in MITuna/tuna/ to a subdirectory, and move the scripts inside MITuna/tuna/miopen/subcmd/ to MITuna/tuna/ ?

@alexandraBara has begun an initiative to add these commonly used tuning commands in the subcmd directory to the miopen_lib. Eventually all of these commands will be invokable by calling for example './go_fish.py miopen import_configs' followed by the arguments for each subcmd. We'll be preserving the scripts ability to be invokable stand-alone as well.

Everything that is miopen related is being relocated to the miopen folder as well. Moving forward, we are considering miopen as one of the user modules to Tuna. So top level Tuna will be limited to the generic functionality of Tuna.

Jenkinsfile
cderb: Those become
tuna.miopen.db.tables
tuna.miopen.utils.config_type
tuna.dbBase.sql_alchemy
tuna.utils.db_utility

Once those are updated we can re-enable it on this branch.
We might should add a version here in the Tuna CI, because this update is going to break past builds.

JehandadKhan (approved):

JehandadKhan:
@msaudulhassan  Should we wait for you to make the changes or can this go in while you work on SolverAnalytics ?

msaudulhassan:
> @msaudulhassan Should we wait for you to make the changes or can this go in while you work on SolverAnalytics ?

@JehandadKhan Let's do this in this PR. The changes are quick to make. I'll push them asap.

Jenkinsfile
msaudulhassan: Updated solver analytics! You may let the solveranalytics tests run now.
Agreed on adding a version number in the CI

msaudulhassan (approved):

PR #849, opened 2023-02-11
Subcommands: import_configs

alexandraBara:
Moving the import_configs.py scripts to run through go_fish.py as a subcommand.
Old way:
```./tuna/import_configs.py -t recurrent_alex_importcfg_7_bn --mark_recurrent -f ../utils/configs/batch_norm.txt -C batch_norm
```
New way:
```./go_fish.py miopen import_configs -t recurrent_alex_importcfg_7_bn --mark_recurrent -f ../utils/configs/batch_norm.txt -C batch_norm```

Notice that this can also happen:
```./go_fish.py -C convolution miopen import_configs -t recurrent_alex_importcfg_7_bn --mark_recurrent -f ../utils/configs/batch_norm.txt -C batch_norm```
since -C is an option for both common scripts and import_config subcommand. In this case, the subcommand value overwrites the common args value.

YAML file/parsing has been updated to support subcommands.

### CODE REVIEW

File: tuna/parse_args.py [hunk 1]
```diff
@@ -153,7 +154,7 @@ def args_check(args, parser):
 
   args.local_machine = not args.remote_machine
 
-  if args.init_session and not args.label:
+  if args.subcommand is not None and args.subcommand == 'init_session' and not args.label:
```
cderb: init_session isn't added as a subcommand.

tuna/miopen/parse_miopen_args.py
```diff
@@ -0,0 +1,78 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+###############################################################################
+"""Module to represent MIOpen subcommands parsers"""
+from tuna.parse_args import TunaArgs, setup_arg_parser
+
+
+def get_import_cfg_parser():
```
cderb: Possible to import this function to import_configs to maintain original functionality?

tuna/miopen/import_configs.py
```diff
@@ -156,87 +99,83 @@ def insert_config(driver, counts, dbt, args):
...
-  LOGGER.info('New configs added: %u', counts['cnt_configs'])
-  if args.tag or args.tag_only:
-    LOGGER.info('Tagged configs: %u', len(counts['cnt_tagged_configs']))
-
+  counts = import_cfgs(args, dbt, logger)
 
-if __name__ == '__main__':
-  main()
+  logger.info('New configs added: %u', counts['cnt_configs'])
+  if args.import_configs.tag or args.import_configs.tag_only:
```
cderb: Not a fan of the layered namespace changes in this file. Could we pass args.import_configs to run_import_configs as args?

tuna/miopen/miopen_lib.py
```diff
@@ -310,6 +340,12 @@ def run(self):
     if self.args.add_tables:
       self.add_tables()
       return None
+
+    if self.args.subcommand is not None and self.args.subcommand == 'import_configs':
+      self.set_import_cfg_batches()
+      run_import_configs(self.args, self.logger)
```
cderb: Could pass self.args.import_configs here.

tuna/miopen/import_configs.py
```diff
@@ -156,87 +99,83 @@ def insert_config(driver, counts, dbt, args):
...
+        parse_line(args, line, counts, dbt, logger)
       except ValueError as err:
-        LOGGER.warning(err)
+        logger.warning(err)
         continue
 
   return counts
 
 
-def main():
```
cderb: If we import the parser from parse_miopen_args.py, I think we can keep main and call run_import_configs after calling parse_args().

tuna/miopen/miopen_lib.py
```diff
@@ -176,6 +188,24 @@ def parse_args(self):
...
 
+  def overwrite_common_args(self):
+    """Overwrite common MIOpen_lib args with subcommand args"""
+    if self.args.subcommand is not None:
+      subc_dict = vars(self.args.get(self.args.subcommand))
+      for sub_key in subc_dict:
+        if sub_key in vars(self.args):
+          self.args[sub_key] = subc_dict.get(sub_key)
+
+  def set_import_cfg_batches(self):
```
cderb: Put next to parsing in parse_miopen_args?

cderb (changes requested):

tuna/miopen/parse_miopen_args.py
alexandraBara: doing that wont maintain the original functionality since the argument parsing has moved.

tuna/parse_args.py
alexandraBara: its not a subcommand, do you want to make it a subcommand?

tuna/miopen/miopen_lib.py
alexandraBara: self.args.import_configs does not contain arguments like: config_type, arch, num_cu etc which are part of the shared args, so the whole thing needs to be passed.

tuna/miopen/import_configs.py
alexandraBara: we would need the parser from miopen_lib that contains the entire scope of the arguments not just the subcommand arguments.

tuna/miopen/miopen_lib.py
alexandraBara: yeah ill move

tuna/miopen/import_configs.py
alexandraBara: Subcommands or subparsers create layers, or levels of arguments. It's the only way we can make sense of so many argument groups. Its a concept we will have to get used to, unless you can come up with a better way to address the issue.

tuna/miopen/import_configs.py
alexandraBara: Since the parser from parse_miopen_args.py gets added as a subcommand, the code in import_configs.py uses it as a subcommand, so instead of args.tag its args.import_configs.tag. In order to call from main we would need to create a parser and add a subcommand to it.

tuna/miopen/import_configs.py
cderb: import_configs is compartmentalized already, it has no knowledge of the other subcommands, so why should it be holding onto their namespaces?
I'm suggesting we constrict to the relevant namespace at input to the function defined here.

tuna/miopen/parse_miopen_args.py
cderb: If we import this function we would additionally need to import set_import_config_batches. Calling the parser returned by this function followed by parser.parse_args() and set_import_config_batches would restore functionality.

tuna/miopen/miopen_lib.py
cderb: Note that get_import_cfg_parser does include config_type and that arch and num_cu are not relevant in this context.
config_type would be present at the submodule namespace level 'self.args.import_configs'.
I went through import_configs again to be sure and args.import_configs universally replaces args in this diff with exception to one location referencing config_type, but that is available in both namespaces. So passing self.args.import_configs would remove the need for this indirection.

tuna/miopen/import_configs.py
cderb: This combines with my assumption that we would pass args.import_configs to the run_import_configs function.

tuna/parse_args.py
cderb: This statement appears to be testing init_session as if it were a subcommand.

cderb (approved):

tuna/miopen/subcmd/import_configs.py
```diff
@@ -106,7 +55,13 @@ def create_query(tag, mark_recurrent, config_id):
   return query_dict
 
 
-def tag_config_v2(driver, counts, dbt, args, new_cf=None):
+def tag_config_v2(
+    driver: Union[DriverConvolution, DriverBatchNorm],
```
cderb: Do type-hints support polymorphism, like declaring with DriverBase?

cderb (approved):

tuna/miopen/subcmd/import_configs.py
alexandraBara: seems like it does, I have updated

reidkwja (approved):

PR #857, opened 2023-02-24
Tuning performance improvements.

cderb:
Update Fin branch
Improve applicability update performance by moving old applicability cleanup to front.
Add session_retry to fin find eval.
Improve get_job and set_job_state functions / queries. Add rollback for both on error and 'skip lock' for 'select for update' query.


Dockerfile
```diff
@@ -174,6 +174,7 @@ ADD tests /tuna/tests/
 ADD utils /tuna/utils/
 ADD requirements.txt /tuna/
 WORKDIR /tuna
+ENV PYTHONPATH=/tuna
```
alexandraBara, line 177: nice

tuna/miopen/worker/fin_class.py
```diff
@@ -309,15 +309,30 @@ def __set_all_configs(self, idx=0, num_blk=1):
...
           master_cfg_list.append(r_dict)
 
-        block_size = len(rows) // num_blk  #size of the config block
-        extra = len(rows) % num_blk  #leftover configs, don't divide evenly
+        #remove old applicability
+        rm_old = ''
+        if self.label:
+          cfg_ids = [str(row.id) for row in rows]
+          cfg_str = ','.join(cfg_ids)
+          rm_old = f"update {self.dbt.solver_app.__tablename__} set applicable=0 where session={self.session_id} and config in ({cfg_str});"
```
alexandraBara: we never actually delete from the applicability table? we just set it to not applicable? this will make this table very very big. Is there a reason to keep old applicability around?

tuna/miopen/worker/fin_class.py
```diff
@@ -142,7 +142,7 @@ def set_db_tables(self):
                               config_type=self.config_type)
 
   def compose_work_objs(self, session, conds):
```
alexandraBara: can you add type hints for this function? and other functions you have touched. Also update the Jenkinsfile to call mypy on these files.

alexandraBara (changes requested):

tuna/miopen/worker/fin_class.py
cderb: I think setting the applicability field was faster than attempting to delete the entry. Anyhow, this only updates the one session where we are creating the present applicability set. I suppose it would be good to remove old sessions and with it data tied to those sessions like applicability, once those tuning sessions are no longer necessary.

alexandraBara (approved):

alexandraBara (approved):

JehandadKhan (approved):

PR #860, opened 2023-03-06
Remove solver_id from session

cderb:
Solver id is tracked by the job table, so it isn't needed in session.


alexandraBara (approved):

reidkwja (approved):

JehandadKhan:
@alexandraBara  Do we need an update on alembic here ?

alexandraBara:
> @alexandraBara Do we need an update on alembic here ?

We are way behind on alembic changes. We stopped using it when it became too much work for little change.

PR #862, opened 2023-03-08
Convolution.py - Type Hint support

technicalgrp89:
tuna/miopen/driver/convolution.py
-- Type Annotations are added
-- checked against mypy tool
-- utils.groovy is updated to run mypy cmd

### DIFF
--- a/tuna/miopen/driver/convolution.py
+++ b/tuna/miopen/driver/convolution.py
@@ -28,7 +28,6 @@
 
 from typing import Dict, Set, Optional
 from re import search
-
 from tuna.utils.logger import setup_logger
 from tuna.miopen.driver.base import DriverBase
 from tuna.miopen.utils.metadata import CONV_CONFIG_COLS
@@ -151,7 +150,7 @@ class DriverConvolution(DriverBase):
     """Compose obj from conv_config row"""
     self.decompose_weight_t(db_obj)
     key: str
-    value: str
+    value: int
     for key, value in db_obj.to_dict(ommit_ts=True, ommit_valid=True).items():
       if key not in ('id', 'input_t', 'weight_t', 'driver'):
         setattr(self, key, value)
@@ -169,9 +168,9 @@ class DriverConvolution(DriverBase):
 
   def get_conv_dict(self) -> dict:
     """Populate c_dict with conv table elems"""
-    c_dict: dict = {}
+    c_dict: Dict[str, int] = {}
     key: str
-    val: str
+    val: int
     for key, val in self.to_dict().items():
       if key in CONV_CONFIG_COLS:
         c_dict[key] = val
@@ -193,7 +192,7 @@ class DriverConvolution(DriverBase):
   def set_defaults(self, defaults) -> None:
     """Set fds defaults"""
     k: str
-    val: str
+    val: int
     for k, val in self.to_dict().items():
       if k in defaults.keys():
         if val is None:
--- a/tuna/miopen/driver/convolution.py
+++ b/tuna/miopen/driver/convolution.py
@@ -25,6 +25,8 @@
 #
 ###############################################################################
 """Module that a convolution MIOpenDriver cmd"""
+
+from typing import Dict, Set, Optional
 from re import search
 
 from tuna.utils.logger import setup_logger
@@ -33,7 +35,8 @@ from tuna.miopen.utils.metadata import CONV_CONFIG_COLS
 from tuna.miopen.utils.helper import get_db_id
 from tuna.miopen.db.miopen_tables import ConvolutionConfig
 from tuna.miopen.utils.metadata import CONV_2D_DEFAULTS, SUPPORTED_CONV_CMDS, PREC_TO_CMD
-from tuna.miopen.utils.metadata import CONV_3D_DEFAULTS, TENSOR_COLS, TABLE_COLS_CONV_MAP, TENSOR_PRECISION
+from tuna.miopen.utils.metadata import CONV_3D_DEFAULTS, TENSOR_COLS
+from tuna.miopen.utils.metadata import TABLE_COLS_CONV_MAP, TENSOR_PRECISION
 from tuna.miopen.utils.metadata import DIRECTION, DIR_MAP, CONV_SKIP_ARGS, INVERS_DIR_MAP
 from tuna.miopen.utils.parsing import get_fd_name, conv_arg_valid, get_fds_from_cmd
 from tuna.miopen.utils.config_type import ConfigType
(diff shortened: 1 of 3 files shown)


tuna/miopen/driver/convolution.py
```diff
@@ -56,37 +63,37 @@ def __init__(self, line=None, cmd=None, db_obj=None, kwargs=None):
...
+    self.in_h: Optional[str] = None
+    self.in_w: Optional[str] = None
+    self.fil_d: Optional[str] = None
+    self.fil_h: Optional[str] = None
+    self.fil_w: Optional[str] = None
+    self.in_channels: Optional[str] = None
+    self.out_channels: Optional[str] = None
+    self.num_dims: Optional[int] = None
+    self.direction: Optional[str] = None
+    self._cmd: Optional[str] = 'conv'
```
alexandraBara: dont think this is optional, should just be str

tuna/miopen/driver/convolution.py
```diff
@@ -100,21 +107,23 @@ def __init__(self, line=None, cmd=None, db_obj=None, kwargs=None):
       self._cmd = cmd
 
   @property
-  def cmd(self):
+  def cmd(self) -> Optional[str]:
```
alexandraBara: also not optional

alexandraBara (changes requested):

tuna/miopen/driver/convolution.py
cderb: There's 3 ways these values could be populated. 
It could be the kwargs from the constructor here. - and so set here immediately
It could be from parsing the driver line. - set by super constructor
It could be from parsing the fdb_key from a db file. - set by super constructor

cderb: Potential solution is to populate these fields with the defaults for conv2d in metadata.

tuna/miopen/driver/convolution.py
```diff
@@ -191,43 +206,43 @@ def set_defaults(self, defaults):
...
     return get_fd_name(tok1, TABLE_COLS_CONV_MAP)
 
   @staticmethod
-  def get_check_valid(tok1, tok2):
+  def get_check_valid(tok1: str, tok2: str) -> str:
     """Check if valid conv arg"""
     return conv_arg_valid(tok1, tok2)
 
-  def get_db_obj(self, keep_id=False):
+  def get_db_obj(self, keep_id: bool = False) -> dict:
```
cderb: ConvolutionConfig is not a dict

cderb (changes requested):

tuna/miopen/driver/convolution.py
technicalgrp89: data set is initialized to CONV_2D_DEFAULTS table from metadata.py

tuna/miopen/driver/convolution.py
```diff
@@ -56,37 +63,37 @@ def __init__(self, line=None, cmd=None, db_obj=None, kwargs=None):
...
+    self.in_d: int = 1
+    self.in_h: int = 32
+    self.in_w: int = 32
+    self.fil_d: int = 1
+    self.fil_h: int = 3
+    self.fil_w: int = 3
+    self.in_channels: int = 3
+    self.out_channels: int = 32
+    self.num_dims: int = 2
+    self.direction: List[str] = ['F', 'B', 'W']
```
cderb: Direction can only be one of these strings.

tuna/miopen/driver/convolution.py
```diff
@@ -191,43 +206,43 @@ def set_defaults(self, defaults):
                          k, self.spatial_dim)
 
   @staticmethod
-  def get_params(tok1):
+  def get_params(tok1: str) -> Callable[[str, str], bool]:
```
cderb: This does not return a callable function. It attempts to return the value returned by get_fd_name. However get_fd_name returns nothing.
Remove this return (leave the function call) and set the return value to none.

tuna/miopen/driver/convolution.py
```diff
@@ -191,43 +206,43 @@ def set_defaults(self, defaults):
...
     """Get full arg name"""
     return get_fd_name(tok1, TABLE_COLS_CONV_MAP)
 
   @staticmethod
-  def get_check_valid(tok1, tok2):
+  def get_check_valid(tok1: str, tok2: str) -> Callable[[str, str], bool]:
```
cderb: This does not return a callable function. It returns the return of conv_arg_valid which is a boolean.

cderb (changes requested):

tuna/miopen/driver/convolution.py
technicalgrp89: retained 'F'

tuna/miopen/driver/convolution.py
technicalgrp89: get_fd_name retruns tok of type str

tuna/miopen/driver/convolution.py
```diff
@@ -134,31 +143,35 @@ def parse_driver_line(self, line):
...
+  def parse_row(self, db_obj: ConvolutionConfig) -> bool:
     """Overwritting base class function for conv"""
     return self.parse_conv_row(db_obj)
 
-  def parse_conv_row(self, db_obj):
+  def parse_conv_row(self, db_obj: ConvolutionConfig) -> bool:
     """Compose obj from conv_config row"""
     self.decompose_weight_t(db_obj)
+    key: str
+    value: str
```
cderb: Refer to init and to_dict, this object has values that may be string or int.

tuna/miopen/driver/convolution.py
```diff
@@ -134,31 +143,35 @@ def parse_driver_line(self, line):
...
 
     return c_dict
 
-  def get_conv_dict(self):
+  def get_conv_dict(self) -> dict:
     """Populate c_dict with conv table elems"""
-    c_dict = {}
+    c_dict: dict = {}
+    key: str
+    val: str
```
cderb: could be string or int

tuna/miopen/driver/convolution.py
```diff
@@ -177,8 +190,10 @@ def config_set_defaults(self):
       self.set_defaults(CONV_2D_DEFAULTS)
       self.num_dims = 2
 
-  def set_defaults(self, defaults):
+  def set_defaults(self, defaults) -> None:
     """Set fds defaults"""
+    k: str
+    val: str
```
cderb: typing

cderb (changes requested):

tuna/miopen/driver/convolution.py
```diff
@@ -125,7 +134,7 @@ def parse_fdb_key(self, line):
     if search(pattern_3d, line):
       setattr(self, 'spatial_dim', 3)
 
-  def parse_driver_line(self, line):
+  def parse_driver_line(self, line: str) -> None:
     super().parse_driver_line(line)
 
     if self.direction and self.direction in DIRECTION:
```
cderb, line 139: Technically self.direction is initialized as an int in the set (1,2,4) by parse_driver_line. This line (140) resolves it to a string.

cderb (approved):

alexandraBara (approved):

PR #864, opened 2023-03-10
perf db layout & backwards input size calculation correction

cderb:
workaround for exporting correct layout
input checking for import_configs to match tensor layouts 
bounding output size calculation to positive number 
re-organizing default values for convolutions


tuna/miopen/subcmd/export_db.py
```diff
@@ -456,6 +457,13 @@ def export_pdb(dbt, args):
       ins_cfg_id = cfg_map[cfg_entry.id]
     else:
       cfg_dict = get_cfg_dict(cfg_entry, cfg_entry.input_t)
+
+      #override cfg layout with fdb key layout
+      if perf_db_entry.fdb_key:
```
alexandraBara, line 462: are we assuming that we imported configs with the wrong layout and so we overwrite it with the fdb key layout?

alexandraBara (approved):

tuna/miopen/subcmd/export_db.py
```diff
@@ -456,6 +457,13 @@ def export_pdb(dbt, args):
...
+        fds, vals, _, _ = parse_pdb_key(perf_db_entry.fdb_key)
+        key_layout = vals[fds.index('out_layout')]
+        cfg_dict['layout'] = key_layout
```
JehandadKhan: This should be an error!

tuna/miopen/subcmd/import_configs.py
```diff
@@ -154,6 +154,10 @@ def parse_line(args: argparse.Namespace, line: str, counts: dict,
     driver = DriverBatchNorm(line, args.command)
   else:
     driver = DriverConvolution(line, args.command)
+    if not driver.out_layout == driver.in_layout == driver.fil_layout:
+      raise ValueError(
+          f"Non-matching layouts {driver.out_layout}, {driver.in_layout}, {driver.fil_layout}"
+      )
```
JehandadKhan: This might not always be true, since the filter layout for NHWC convolution is CNHW ?

JehandadKhan: Lets confirm by running some driver commands

JehandadKhan (changes requested):

tuna/miopen/subcmd/export_db.py
cderb: There was a fin issue which was causing layouts other than the one specified by the config to be tuned.
changing this statement to raise an error instead.

reidkwja: [codebase design question] why do we need to override cfg layout with fdb keys?

alexandraBara (approved):

JehandadKhan (approved):

PR #865, opened 2023-03-10
flask.py -  typehint support

technicalgrp89:
tuna/flask_example.py - type hints support added & mypy compatiblity
tuna/flask.py - un used file removed from the tuna repo.
vars/utils.groovy - mypy cmd support.


alexandraBara (approved):

tuna/flask_example.py
```diff
@@ -50,9 +51,10 @@ def get_table_example(grafana_req, data):
   EXAMPLE_TABLE['rows'].append(['val4', 'ex4', '4', '1.08'])
 
   #To populate the table with data from your DB:
+  res: str
   with DbSession() as session:
-    query = session.query(ConvolutionFindDB.valid,
-                          ConvolutionFindDB.kernel_time).limit(5).all()
+    query: str = session.query(ConvolutionFindDB.valid,
```
cderb: Not a str. Will have to consult sqlalchemy docs for actual typing, but acts like a list.

cderb (changes requested):

tuna/flask_example.py
alexandraBara: @technicalgrp89  i believe its a list of DB rows. You can check the API of SQLAlchemy here: https://docs.sqlalchemy.org/en/14/core/connections.html#sqlalchemy.engine.Result.all if you search for '.all()' on this page you might find more details

technicalgrp89: changed to list

cderb (approved):

PR #867, opened 2023-03-14
Catch layout mismatch

cderb:
add layout check after fin job
increase jobs in queue


tuna/worker_interface.py
```diff
@@ -157,9 +157,14 @@ def compose_work_objs(self, session: DbSession,
     cond_str = ' AND '.join(conds)
     if cond_str:
       cond_str = f"WHERE {cond_str}"
-    cond_str += f" ORDER BY retries ASC LIMIT {self.claim_num} FOR UPDATE SKIP LOCKED"
+    cond_str += f" ORDER BY retries ASC LIMIT {self.claim_num} FOR UPDATE"
+    #try once without waiting for lock
```
alexandraBara, line 161: whats the reasoning for this?

cderb: I've changed this to first do with "SKIP LOCKED" and if it fails to gather any results, default to without. Always running with "SKIP LOCKED" would sometimes fail to gather waiting jobs.

tuna/miopen/worker/fin_class.py
```diff
@@ -706,6 +723,7 @@ def __update_fdb_w_kernels(self,
         if fdb_obj[check_str]:
           #returned entry is added to the table
           fdb_entry = self.__compose_fdb_entry(session, fin_json, fdb_obj)
+          self.__check_layout_mismatch(fdb_entry, slv_stat)
```
alexandraBara, line 726: so if the layout doesnt match, this wont kill the tuning job it will just set the error status for the job. Should we at least print some warning?

cderb: Yes, this will set the job state to errored with an appropriate result message. There's no need to stop the tuning overall.

alexandraBara (approved):

reidkwja (approved):

PR #869, opened 2023-03-17
tuna/go_fish.py

technicalgrp89:
Type Annotation support to the present code base
mypy support.

### DIFF
--- a/tuna/go_fish.py
+++ b/tuna/go_fish.py
@@ -25,19 +25,25 @@
 #
 ###############################################################################
 """! @brief Script to launch tuning jobs, or execute commands on available machines"""
+
 import argparse
 import sys
+import logging
+from typing import Dict, List, Any, Union
 from tuna.utils.logger import setup_logger
 from tuna.libraries import Library
 from tuna.lib_utils import get_library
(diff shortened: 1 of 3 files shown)


tuna/go_fish.py
```diff
@@ -60,20 +66,25 @@ def parse_args():
...
+  library: Union[Example, MIOpen]
+  yaml_files: List[str]
   library = get_library(args)
   yaml_files = [args['yaml']]
 
   #case with yaml file
   if args['yaml']:
     yaml_files = parse_yaml(args['yaml'], args['lib'])
 
+  worker_lst: list
```
alexandraBara, line 87: what is this type? i dont see this imported anywhere, i see 'List' but not list

alexandraBara (changes requested):

tuna/go_fish.py
technicalgrp89: list is a built in generic type( python 3.9) . List is defined in typing lib, which I hav not used here.

alexandraBara (approved):

cderb (approved):

PR #873, opened 2023-03-30
Add flag for limiting GPUs

cderb:
Handy for debug, or for adapting to configurations with memory issues.

### DIFF
--- a/tuna/miopen/miopen_lib.py
+++ b/tuna/miopen/miopen_lib.py
@@ -100,6 +100,12 @@ class MIOpen(MITunaInterface):
                         dest='reset_interval',
                         required=False,
                         help='Restart interval for job in hours.')
+    parser.add_argument(
+        '--gpu_lim',
+        dest='gpu_lim',
+        type=int,
+        default=None,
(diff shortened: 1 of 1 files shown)


alexandraBara (approved):

tuna/miopen/miopen_lib.py
```diff
@@ -100,6 +100,12 @@ def parse_args(self):
                         dest='reset_interval',
                         required=False,
                         help='Restart interval for job in hours.')
+    parser.add_argument(
+        '--gpu_lim',
+        dest='gpu_lim',
+        type=int,
+        default=None,
```
reidkwja, line 107: @cderb Would having a default number instead of None be a good alternative?

reidkwja (approved):

tuna/miopen/miopen_lib.py
cderb: In most cases we wouldn't want to impose a cap on the number of GPUs.
Though I suppose if the default was set to the integer equivalent of  infinite, then the if statement later would not have to check for None.

PR #874, opened 2023-03-30
Reduce tensor lookup time.

cderb:
cache tensor table to improve lookup time in driver
correct default layout values

### DIFF
--- a/tuna/utils/db_utility.py
+++ b/tuna/utils/db_utility.py
@@ -244,6 +244,26 @@ def get_class_by_tablename(tablename):
       return c
 
 
+def build_dict_val_key(obj, exclude=['id']):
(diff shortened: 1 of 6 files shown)


tuna/utils/db_utility.py
```diff
@@ -244,6 +244,26 @@ def get_class_by_tablename(tablename):
       return c
 
 
+def build_dict_val_key(obj, exclude=['id']):
```
alexandraBara: these need to be unit tested

alexandraBara (changes requested):

JehandadKhan (approved):

alexandraBara (approved):

technicalgrp89 (approved):

PR #875, opened 2023-03-30
Remove redundant branch from import config

cderb:
Deleted else branch is the same as function tag_config_v2

### DIFF
--- a/tuna/miopen/subcmd/import_configs.py
+++ b/tuna/miopen/subcmd/import_configs.py
@@ -115,18 +115,7 @@ def insert_config(driver: DriverBase, counts: dict, dbt: MIOpenDBTables,
       except IntegrityError as err:
         logger.warning("Err occurred: %s", err)
         session.rollback()
-    else:
-      try:
-        if args.mark_recurrent or args.tag:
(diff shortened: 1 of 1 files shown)


tuna/miopen/subcmd/import_configs.py
```diff
@@ -115,18 +115,7 @@ def insert_config(driver: DriverBase, counts: dict, dbt: MIOpenDBTables,
       except IntegrityError as err:
         logger.warning("Err occurred: %s", err)
         session.rollback()
-    else:
-      try:
-        if args.mark_recurrent or args.tag:
```
alexandraBara, line 120: this is the only place we are using the recurrent argument. If this goes, then we should remove the recurrent argument and col in the DB as well. I dont think we have used it in a long time. Otherwise this leaves us with more dead code

cderb: check line 130 / 119

JehandadKhan (approved):

alexandraBara (approved):

reidkwja (approved):

PR #877, opened 2023-03-31
Integrating load_job subcmd into MIOpen lib go fish  

reidkwja:
This PR would integrate the load_job subcmd into MIOpen to run as the following: 
./tuna/go_fish.py miopen --load_job --session_id 1 --label some_label [..other args]


vars/utils.groovy
```diff
@@ -171,7 +171,8 @@ def finFindCompile(){
         println "Count(*) conv_config table: ${num_cfg}"
         runsql("delete from conv_job;")
         runsql("alter table conv_job AUTO_INCREMENT=1;")
-        sh "./tuna/miopen/subcmd/load_job.py -l finFind_${branch_id} --all_configs --fin_steps \"miopen_find_compile, miopen_find_eval\" --session_id ${sesh1} ${job_lim}"
+        sh "./tuna/go_fish.py miopen load_job -l finFind_${branch_id} --all_configs --fin_steps \"miopen_find_compile,miopen_find_eval\" --session_id ${sesh1} ${job_lim}"
+        //sh "./tuna/miopen/subcmd/load_job.py -l finFind_${branch_id} --all_configs --fin_steps \"miopen_find_compile, miopen_find_eval\" --session_id ${sesh1} ${job_lim}"
```
cderb: Clean up comments like this before merging.

tuna/miopen/miopen_lib.py
```diff
@@ -163,7 +170,7 @@ def parse_args(self):
       print_solvers()
       raise ValueError('Printing solvers...')
 
-    if self.args.fin_steps:
+    if self.args.fin_steps and self.args.subcommand != 'load_job':
```
alexandraBara, line 178:
```suggestion
    if self.args.fin_steps and self.args.subcommand and  self.args.subcommand != 'load_job':
```
you need to check if there is a subcommand first before u check the name otherwise u rist self.args.subcommand is None

tuna/miopen/yaml_files/sample1.yaml
```diff
@@ -26,6 +26,9 @@ import_configs:
   framework: 'Pytorch'
   fw_version: 1
 
+load_job:
+  enabled: true
```
alexandraBara, line 30: have you tested this? i think there are some other arguments required in load_job like...reason and such, maybe some fin steps

tests/test_yaml_parser.py
```diff
@@ -52,7 +53,7 @@ def test_yaml_parser():
 
 def parse_miopen_yaml1(miopen_yaml, miopen):
   yaml_files = parse_yaml(miopen_yaml, miopen)
-  assert len(yaml_files) == 3
+  assert len(yaml_files) == 4
```
alexandraBara, line 76: add the 4 dict below and compare expected and actual as its done with the previous 3 dicts. To find what that dict looks like u will have to run go_fish.py with a load_job yaml file. Also please add a load_job yaml file to /MITuna/tuna/miopen/yaml_files call it sample4.yaml, this should obviously be a load_jobs yaml

alexandraBara (changes requested):

tests/test_yaml_parser.py
alexandraBara: currently you can run go_fish will all 3 yaml files from tuna/miopen/yaml_files like:
```
./go_fish.py miopen --yaml miopen/yaml_files/sample1.yaml 
```

tuna/miopen/miopen_lib.py
reidkwja: It would error out under that case

tuna/miopen/subcmd/load_job.py
```diff
@@ -40,87 +42,21 @@
...
-  args = parser.parse_args()
 
+def arg_fin_steps(args: argparse.Namespace):
+  """fin steps for load jobs"""
   if args.fin_steps:
     steps = [x.strip() for x in args.fin_steps.split(',')]
     args.fin_steps = set(steps)
 
+
+def arg_solvers(
```
alexandraBara, line 60: this function is not called, this causes a bug in that your PR does not take into account when -A or -s is passed to load_job. 
For instance, if you run this cmd in develop:
```
./tuna/miopen/subcmd/load_job.py -t recurrent_rk_prase_jobs_61 -l recurrent_rk_prase_jobs_61 --session_id 1 -A miopenConvolutionAlgoGEMM
```
this is what args looks like:
```
./tuna/miopen/subcmd/load_job.py -t recurrent_rk_prase_jobs_61 -l recurrent_rk_prase_jobs_61 --session_id 1 -A miopenConvolutionAlgoGEMM
Namespace(algo='miopenConvolutionAlgoGEMM', all_configs=False, cmd=None, config_type=<ConfigType.convolution: 'convolution'>, fin_steps=None, label='recurrent_rk_prase_jobs_61', only_dynamic=False, session_id=1, solvers=[('GemmFwd1x1_0_1', 88), ('GemmFwd1x1_0_1_int8', 89), ('GemmFwd1x1_0_2', 90), ('GemmFwdRest', 91), ('GemmBwd1x1_stride2', 95), ('GemmBwd1x1_stride1', 96), ('GemmBwdRest', 97), ('GemmWrw1x1_stride1', 101), ('GemmWrwUniversal', 102)], tag='recurrent_rk_prase_jobs_61', tunable=False, yaml=None)
```
As you can see it picked up the solvers for that -A algo.
Now, if you run the same command in your PR, this is what args looks like:
```
 ./tuna/go_fish.py miopen load_job -t recurrent_rk_prase_jobs_61 -l recurrent_rk_prase_jobs_61 --session_id 1 -A miopenConvolutionAlgoGEMM
2023-04-04 14:33:16,997 - go_fish - INFO -  [go_fish.py:71] - ['./tuna/go_fish.py', 'miopen', 'load_job', '-t', 'recurrent_rk_prase_jobs_61', '-l', 'recurrent_rk_prase_jobs_61', '--session_id', '1', '-A', 'miopenConvolutionAlgoGEMM']
2023-04-04 14:33:16,997 - go_fish - INFO -  [go_fish.py:72] - 11
Namespace(algo='miopenConvolutionAlgoGEMM', all_configs=False, cmd=None, config_type=<ConfigType.convolution: 'convolution'>, fin_steps=None, label='recurrent_rk_prase_jobs_61', only_dynamic=False, session_id=1, solvers=None, tag='recurrent_rk_prase_jobs_61', tunable=False, yaml=None)

```
As you can see, the args results from this run do not contain any solvers, that is because the arg_solver function is never called. 
Please address

alexandraBara (changes requested):

tuna/miopen/subcmd/load_job.py
alexandraBara: @reidkwja  given that this was indeed the issue that was causing bad json file and hence an miopen segfault, please update the load_job unit test and look for:

1. correct number of jobs being set for the expected solvers in a given algo, specified with -A
2. correct solver being set for a job that specifies a solver.
Currently the load job utest only checks that jobs are loaded but not how many

alexandraBara (changes requested):
FYI, the subcmd dir is not linted, please add a lint call in vars/groovy.py and lint load_jobs.py and address the issues

tuna/miopen/subcmd/load_job.py
```diff
@@ -47,7 +49,10 @@
 
 def arg_fin_steps(args: argparse.Namespace):
   """fin steps for load jobs"""
-  if args.fin_steps:
+  if args.fin_steps == '':
```
alexandraBara, line 52: this should fail in the parse_args function. --fin_steps expects a value after. Please double check by running something like:
go_fish.py miopen load_job --fin_steps -s SomeSolverName [..other required args].
Should error out with something like: --fin_steps expects a value

tuna/miopen/subcmd/load_job.py
```diff
@@ -207,15 +151,15 @@ def add_jobs(args, dbt):
...
 
     fin_step_str = 'not_fin'
     if args.fin_steps:
       fin_step_str = ','.join(args.fin_steps)
     query = f"select config, solver from {dbt.job_table.__tablename__} where session={args.session_id} and fin_step='{fin_step_str}' and reason='{args.label}'"
-    LOGGER.info(query)
+    logger.info(query)
     ret = session.execute(query)
-    pre_ex = {}
+    pre_ex: Dict[str, Dict[str, bool]] = {}
```
alexandraBara, line 161: is the value in this dict str: bool or str: int. Im thinking it would map solver name to solver_id. @cderb  can you confirm?

tuna/miopen/subcmd/load_job.py
```diff
@@ -246,38 +190,46 @@ def add_jobs(args, dbt):
...
       break
 
   return counts
 
 
-def main():
-  """ main """
-  args = parse_args()
+def run_load_job(args: argparse.Namespace, logger: logging.Logger):
+  """trigger the script run in miopen lib"""
```
alexandraBara:
```suggestion
  """Load jobs based on cmd line arguments"""
```

tuna/miopen/subcmd/load_job.py
```diff
@@ -246,38 +190,46 @@ def add_jobs(args, dbt):
...
     except ValueError as terr:
-      LOGGER.error(terr)
-
-  cnt = add_jobs(args, dbt)
+      logger.error(terr)
+  if args.solvers or args.algo:
+    args = arg_solvers(args, logger)
 
+  cnt = add_jobs(args, dbt, logger)
   print(f"New jobs added: {cnt}")
```
alexandraBara, line 223: Make this also return the number of jobs it added, then you can use that to check the expected number in the utest

alexandraBara (changes requested):

tuna/miopen/subcmd/load_job.py
reidkwja: Could this act as a secondary check from a bottom-up hierarchy?

tuna/miopen/yaml_files/sample1.yaml
reidkwja: tested this file and its working so far

alexandraBara (approved):

tuna/miopen/subcmd/load_job.py
cderb: If we observe this line: pre_ex[config][solver] = True
I would say it is a Dict[int, Dict[str, bool]], as config is an id and solver is a solver namestring. Though it isn't a huge deal if config is compared here as a string.

tuna/miopen/yaml_files/sample1.yaml
cderb: label is a required field for load_jobs, but up in the arg parsing it has a default value of 'new', perhaps that default should be removed and a label added here in the sample.yaml

cderb (approved):

PR #880, opened 2023-04-04
Populate driver col in conv_config

alexandraBara:
Populating driver col in conv_config + restored pylint for import_configs


tuna/miopen/driver/convolution.py
```diff
@@ -168,7 +168,7 @@ def compose_tensors(self, keep_id: bool = False) -> dict:
 
   def get_conv_dict(self) -> dict:
     """Populate c_dict with conv table elems"""
-    c_dict: Dict[str, int] = {}
+    c_dict: Dict[str, Any] = {}
```
technicalgrp89, line 171: What is the reason to define it as Any from integer? reason?

tuna/miopen/driver/convolution.py
```diff
@@ -177,6 +177,7 @@ def get_conv_dict(self) -> dict:
 
     c_dict['input_tensor'] = super().get_input_t_id()
     c_dict['weight_tensor'] = super().get_weight_t_id()
+    c_dict['driver'] = str(self)
```
technicalgrp89, line 180: whats the actual value we get here as self object?

tuna/miopen/driver/convolution.py
alexandraBara: its got  a string as a value now as well, they are not all integers anymore

technicalgrp89: ok

technicalgrp89 (approved):

tuna/miopen/subcmd/import_configs.py
```diff
@@ -180,7 +180,8 @@ def import_cfgs(args: argparse.Namespace, dbt: MIOpenDBTables,
         parse_line(args, line, counts, dbt, logger)
       except ValueError as err:
         logger.warning(err)
-        continue
+        if 'Skipping line' in str(err):
+          continue
```
cderb: Continue looks like  a null op here.

tuna/miopen/utils/metadata.py
```diff
@@ -220,6 +220,7 @@
     'FP16': 'FP16',
     'convfp16': 'FP16',
     'BF16': 'BF16',
+    'BFP16': 'BF16',
```
cderb: Is it anticipated that Tuna encounters BFP16 as input? I thought it was intended to eliminate this.

cderb (approved):

PR #881, opened 2023-04-06
Sp/tuna machine typehint

technicalgrp89:
machine.py - code is made compatible with mypy stds. Varables/functions/input/output/retrun parameters data types are defined.


tuna/machine.py
```diff
@@ -283,31 +306,48 @@ def get_properties(self):
...
+    fout_2: BufferedWriter
+
+    t_filename: Union[str, 'os.PathLike[Any]']
     if is_temp:
       assert filename is None
-      _, filename = tempfile.mkstemp()
-    else:
-      assert filename is not None
+      _, t_filename = tempfile.mkstemp()
+      return t_filename
```
alexandraBara: if this returns here, the contents never gets written to the json, hence and empty json file and FIN will complain the json is not as expected. Please address by removing the return statement here

tuna/machine.py
```diff
@@ -283,31 +306,55 @@ def get_properties(self):
...
+  def write_file(self, contents: bytes, filename: Union[str, bytes, 'os.PathLike[str]',\
+  None] = None, is_temp: bool = False) -> Union[str, bytes, 'os.PathLike[str]', None]:
     """
     Write a file to this machine containing contents
     """
+    cnx: Connection
+    ftp: Optional[paramiko.sftp_client.SFTPClient] = None
+    fout: SFTPFile
+    fout_2: IO[Any]
+    t_file: bool = False
```
alexandraBara: this is not needed, 1 variable is enough to hold the temp file. You should just return filename

tuna/machine.py
```diff
@@ -283,31 +306,55 @@ def get_properties(self):
...
-      ftp = cnx.ssh.open_sftp()
-      with ftp.open(filename, 'wb') as fout:
-        fout.write(contents)
-        fout.flush()
+      if cnx is not None and cnx.ssh is not None:
+        ftp = cnx.ssh.open_sftp()
+      if ftp is not None:
+        with ftp.open(filename, 'wb') as fout:  #type: ignore
+          fout.write(contents)
+          fout.flush()
```
alexandraBara: this is why you get the error, this code is not reached, you have changed the way the code works here, please undo

alexandraBara (changes requested):

tuna/machine.py
```diff
@@ -283,31 +307,46 @@ def get_properties(self):
...
+
+    #t_filename: Union[str, 'os.PathLike[str]']
     if is_temp:
       assert filename is None
       _, filename = tempfile.mkstemp()
     else:
       assert filename is not None
-    if self.local_machine:  # pylint: disable=no-member ; false alarm
+
+    if self.local_machine and filename is not None:  # pylint: disable=no-member ; false alarm
```
alexandraBara: i dont think the 'filename is not none' part is required here. The filename is established beforehand or passed in as not none.

tuna/machine.py
```diff
@@ -283,31 +307,46 @@ def get_properties(self):
...
-  def write_file(self, contents, filename=None, is_temp=False):
+  def write_file(self, contents: bytes, filename: Union[str, bytes, 'os.PathLike[str]',\
+  None] = None, is_temp: bool = False) -> Union[str, bytes, 'os.PathLike[str]', None]:
     """
     Write a file to this machine containing contents
     """
+    cnx: Connection
+    ftp: Optional[paramiko.sftp_client.SFTPClient] = None
+
+    #t_filename: Union[str, 'os.PathLike[str]']
```
alexandraBara: remove

tuna/machine.py
```diff
@@ -319,21 +358,28 @@ def read_file(self, filename, byteread=False):
       ftp.getfo(filename, content_io)
       ret = content_io.getvalue()
       if not byteread:
-        ret = ret.decode()
+        ret_str = ret.decode()
+        return ret_str
```
alexandraBara: can you set ret as Any[str, bytes] and return ret, thus removing the extra ret_str and leave the code as it was to just return ret.

tuna/machine.py
```diff
@@ -377,25 +432,15 @@ def get_gpu_clock(self, gpu_num=0):
...
-      mclk_cmd = f'{ROCMSMI} --setmclk {self.mclk}'
-      cnx.exec_command(mclk_cmd)
-
-  def restart_server(self, wait=True):
+  def restart_server(self, wait: bool = True) -> bool:
     """restart the machine"""
     logger = self.get_logger()
 
     logger.warning('Sending remote reboot command')
     if self.ipmi_ip is not None and not self.ipmi_inaccessible:
```
alexandraBara:
```suggestion
    if self.ipmi_ip is not None and not self.ipmi_inaccessible and self.mmi:
```

tuna/machine.py
```diff
@@ -377,25 +432,15 @@ def get_gpu_clock(self, gpu_num=0):
...
       logger.info('Using IPMI to reboot machine')
-      self.mmi.restart_server()
+      if self.mmi is not None:
```
alexandraBara: remove this condition and expand the above one like i suggested

alexandraBara (approved):

alexandraBara:
sorry clicked 'Approved' by mistake, i meant to do 'Request changes' alongside my comments :)

tuna/machine.py
```diff
@@ -186,22 +207,22 @@ def get_gpu(self, idx):
...
 
-  def get_cpus(self):
-    """return cpu details array"""
-    if not self.cpus:
-      self.get_properties()
-
-    return self.cpus
-
-  def parse_agents(self):  #pylint: disable=too-many-locals
+  def parse_agents(self) -> dict:  #pylint: disable=too-many-locals
```
cderb, line 209: shouldn't disable pylint

tuna/machine.py
```diff
@@ -83,63 +93,73 @@ def __init__(self, **kwargs):
...
-    self.log_list = {}
-    self.num_cpus = 0
+    self.cnx_list: dict = {}
+    self.log_list: dict = {}
+    self.num_cpus: int = 0
+    self.avail_gpus: List[Dict[int, str]]
+    self.sclk: int
+    self.mclk: int
+    self.gpus: List[Dict[Any, Any]]
+    self.cpus: List[dict]
```
cderb: inconsistent with get_properties return type.

tuna/machine.py
```diff
@@ -283,31 +307,46 @@ def get_properties(self):
 
     return self.cpus, self.gpus
 
-  def write_file(self, contents, filename=None, is_temp=False):
+  def write_file(self, contents: bytes, filename: Union[str, bytes, 'os.PathLike[str]',\
```
cderb: the filename as an input to this function should only be a string

tuna/machine.py
```diff
@@ -347,14 +393,23 @@ def exec_command(self, command, timeout=LOG_TIMEOUT):
...
+    stdout = Optional[IO[str]]
+    gpu_clk_cmd: str = f'{ROCMSMI} -c'
     _, stdout, _ = self.connect().exec_command(gpu_clk_cmd)
 
-    base_idx = None
-    gpu_clk = []
+    base_idx: Union[int, None] = None
+    gpu_clk: List[Dict[str, int]] = []
+    gpu_idx: int = 0
+    level: Union[int, str]
```
cderb: appears to be int only

tuna/machine.py
```diff
@@ -347,14 +393,23 @@ def exec_command(self, command, timeout=LOG_TIMEOUT):
...
+    sclk: Union[SupportsIndex, slice]
+    mclk: Union[SupportsIndex, slice]
```
cderb: sclk and mclk are clock levels, level is an int.

tuna/machine.py
```diff
@@ -408,12 +453,14 @@ def restart_server(self, wait=True):
...
+  def chk_gpu_status(self, gpu_id: int) -> bool:
     """check gpu status, can clinfo find the device"""
-    logger = self.get_logger()
+    line: str
+    stdout: ChannelFile
+    logger: logging.Logger = self.get_logger()
     cnx = self.connect()
 
-    if gpu_id not in self.avail_gpus:
+    if gpu_id not in self.avail_gpus:  #type: ignore
```
cderb: type ignore?

tuna/machine.py
technicalgrp89: ok

tuna/machine.py
technicalgrp89: ok. filename input changed to str type

tuna/machine.py
technicalgrp89: ok

tuna/machine.py
alexandraBara: this should be addressed in a different PR

tuna/machine.py
technicalgrp89: I used Union[str, Bytes].

tuna/machine.py
technicalgrp89: check added for self.mmi

tuna/machine.py
technicalgrp89: check is removed and added in line 440

tuna/machine.py
technicalgrp89: self.mmi is not None added

tuna/machine.py
technicalgrp89: I havnot edited nor changed this condition it is retained as it is.

tuna/machine.py
technicalgrp89: corrected.  def get_properties(self): -> Tuple[List[Dict], List[Dict[Any, Any]]]

tuna/machine.py
technicalgrp89: Level type changed to int

tuna/machine.py
technicalgrp89: variable type correctd.

tuna/machine.py
technicalgrp89: removed.

alexandraBara (approved):

tuna/machine.py
```diff
@@ -43,37 +50,39 @@
...
+  __tablename__: str = "machine"
+  hostname: str = Column(Text, nullable=False)
+  port: int = Column(INTEGER(11, unsigned=True),
+                     nullable=False,
+                     server_default="22")
+  local_ip: str = Column(Text, nullable=True)
+  local_port: int = Column(INTEGER, server_default="22")
+  user: str = Column(Text, nullable=False)
+  password: str = Column(Text, nullable=False)
+  avail_gpus: List[Dict[int, str]] = Column(Text, nullable=False)
```
cderb: This is a List[int], see line 119.

tuna/machine.py
```diff
@@ -83,63 +92,73 @@ def __init__(self, **kwargs):
     self.__dict__.update(
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
 
-    self.cnx_list = {}
-    self.log_list = {}
-    self.num_cpus = 0
+    self.cnx_list: dict = {}
+    self.log_list: dict = {}
+    self.num_cpus: int = 0
+    self.avail_gpus: List[Dict[int, str]]
```
cderb: wrong type

tuna/machine.py
```diff
@@ -83,63 +92,73 @@ def __init__(self, **kwargs):
...
       self.logger = setup_logger('Machine_{self.id}')
       self.mmi = MachineManagementInterface(self.ipmi_ip, self.ipmi_port,
                                             self.ipmi_user, self.ipmi_password)
 
       if not self.avail_gpus is None:
-        self.avail_gpus = [int(val) for val in self.avail_gpus.split(',')]
+        self.avail_gpus = [
+            int(val) for val in self.avail_gpus.split(',')  #type: ignore
+        ]  #type: ignore
         self.num_gpus = len(self.avail_gpus)
```
cderb, line 136: avail_gpus could come in through kwargs as a str. So it's ok to ignore that type complication here. This input parsing should probably be moved outside of this constructor in a separate PR.

tuna/machine.py
```diff
@@ -83,63 +92,73 @@ def __init__(self, **kwargs):
...
-      self.id = 0  # pylint: disable=invalid-name
+
+      self.id: int = 0  # pylint: disable=invalid-name
       self.get_properties()
 
       if self.gpus:
         self.arch = self.gpus[0]['arch']
         self.num_cu = self.gpus[0]['num_cu']
-        self.avail_gpus = list(range(len(self.gpus)))
+        self.avail_gpus = list(range(len(self.gpus)))  #type: ignore
```
cderb: no need to ignore type here

tuna/machine.py
```diff
@@ -161,23 +180,24 @@ def connect(self, abort=None):
...
+  def get_num_cpus(self) -> int:
     """return number of available cpus"""
+    stdout: TextIO
     _, stdout, _ = self.connect().exec_command('nproc')
     self.num_cpus = int(stdout.readline())
 
     return self.num_cpus
 
-  def get_avail_gpus(self):
+  def get_avail_gpus(self) -> List[Dict[int, str]]:
```
cderb: wrong type

tuna/machine.py
```diff
@@ -161,23 +180,24 @@ def connect(self, abort=None):
...
     """return list of available gpus"""
     if not self.avail_gpus:
       if not self.gpus:
         self.get_properties()
-        self.avail_gpus = range(len(self.gpus))
+        self.avail_gpus = range(len(self.gpus))  #type: ignore
```
cderb: don't ignore type.

cderb (changes requested):

tuna/machine.py
technicalgrp89: changed.

tuna/machine.py
technicalgrp89: changed

tuna/machine.py
technicalgrp89: Ok. separate PR is ideal.

tuna/machine.py
technicalgrp89: removed

tuna/machine.py
technicalgrp89: error: Incompatible types in assignment (expression has type "range", variable has type "List[int]") 
used list() o enable the same

tuna/machine.py
technicalgrp89: changed to List[int]

tuna/machine.py
cderb: List[int]

tuna/machine.py
cderb: Dict[Any, Any] = dict, please use most succinct notation.
The keys for these dicts are always str, so Dict[str, Any] would be the most specific.

tuna/machine.py
```diff
@@ -161,23 +180,24 @@ def connect(self, abort=None):
...
     if not self.avail_gpus:
       if not self.gpus:
         self.get_properties()
-        self.avail_gpus = range(len(self.gpus))
+        self.avail_gpus = list(range(len(self.gpus)))
         self.num_gpus = len(self.avail_gpus)
     return self.avail_gpus
 
-  def get_gpu(self, idx):
+  def get_gpu(self, idx: int) -> Optional[Dict[int, str]]:
```
cderb: Optional[Dict[str, Any]]

cderb (changes requested):

tuna/machine.py
technicalgrp89: Dict definitions are changed

tuna/machine.py
```diff
@@ -322,18 +358,24 @@ def read_file(self, filename, byteread=False):
...
 
-  def make_temp_file(self):
+  def make_temp_file(self) -> Union[str, Text]:
     """
     Make an empty temp file on this machine
     """
     return self.write_file(b'', is_temp=True)
 
-  def exec_command(self, command, timeout=LOG_TIMEOUT):
+  def exec_command(self, command: str, timeout: int = LOG_TIMEOUT) -> Tuple[int, str,\
```
cderb: Return value is the same as Connection.exec_command
Which is Tuple[int, StringIO, StringIO]
Remove line 387+388 because err returned from call is a StringIO and not a ChannelFile.

cderb (changes requested):

tuna/machine.py
technicalgrp89: 387,388 removed. 
Line number 395: function get_gpu_clock, if stdout is null case added, as return False.

cderb: out is not a str.

tuna/machine.py
```diff
@@ -408,9 +448,11 @@ def restart_server(self, wait=True):
       return True
     return False
 
-  def chk_gpu_status(self, gpu_id):
+  def chk_gpu_status(self, gpu_id: int) -> bool:
     """check gpu status, can clinfo find the device"""
-    logger = self.get_logger()
+    line: str
+    stdout: ChannelFile
```
cderb: cnx.exec_command no longer returns stdout as a channel file, stdout is a StringIO. Remove lines that treat it as a ChannelFile.

tuna/machine.py
```diff
@@ -442,14 +484,19 @@ def chk_gpu_status(self, gpu_id):
...
-  def getusedspace(self):
+  def getusedspace(self) -> float:
     """examine used space on disk"""
-    logger = self.get_logger()
+    logger: logging.Logger = self.get_logger()
     logger.info("Getting free space")
+    lst: list
+    per: float
+    output: bytes
+    ssh_stdout: ChannelFile
```
cderb: ssh_stdout is StringIO.

cderb (changes requested):

tuna/machine.py
technicalgrp89: if hasattr(stdout, 'channel'):
      stdout.channel.settimeout(30) , this ;ine leads to machine.py:467: error: "StringIO" has no attribute "channel"

technicalgrp89: an we remove this check/. or we can do it on err output, as done in different classes.  if err is not None and hasattr(err, 'channel'): as in worker interface. Let me know

tuna/machine.py
technicalgrp89: changed

tuna/machine.py
technicalgrp89: changed

tuna/machine.py
cderb: remove this check, as StringIO does not have this attribute.

cderb (changes requested):
Incorrect types still present around cnx.exec_command calls

tuna/machine.py
technicalgrp89: check removed.

tuna/machine.py
```diff
@@ -322,18 +358,24 @@ def read_file(self, filename, byteread=False):
...
+  def exec_command(self, command: str, timeout: int = LOG_TIMEOUT) -> Tuple[int, StringIO,\
+  StringIO]:
     """
     Execute a command on this machine
     - through docker if on a remote machine
     - no docker on local machine
     """
+    cnx: Connection
+    ret_code: int
+    out: StringIO = StringIO()
```
cderb, line 376: out: StringIO
is sufficient

tuna/machine.py
```diff
@@ -342,19 +384,27 @@ def exec_command(self, command, timeout=LOG_TIMEOUT):
...
-      err.channel.settimeout(LOG_TIMEOUT)
 
     return ret_code, out, err
 
-  def get_gpu_clock(self, gpu_num=0):
+  def get_gpu_clock(self, gpu_num: int = 0) -> Union[Tuple[int, int], bool]:
     """query gpu clock levels with rocm-smi"""
-
-    gpu_clk_cmd = f'{ROCMSMI} -c'
+    stdout: StringIO = StringIO()
```
cderb, line 392: stdout: StringIO
is sufficient

tuna/machine.py
```diff
@@ -408,9 +447,11 @@ def restart_server(self, wait=True):
       return True
     return False
 
-  def chk_gpu_status(self, gpu_id):
+  def chk_gpu_status(self, gpu_id: int) -> bool:
     """check gpu status, can clinfo find the device"""
-    logger = self.get_logger()
+    line: str
+    stdout: StringIO = StringIO()
```
cderb, line 453: stdout: StringIO
is sufficient

tuna/machine.py
```diff
@@ -442,14 +481,19 @@ def chk_gpu_status(self, gpu_id):
...
-  def getusedspace(self):
+  def getusedspace(self) -> float:
     """examine used space on disk"""
-    logger = self.get_logger()
+    logger: logging.Logger = self.get_logger()
     logger.info("Getting free space")
+    lst: list
+    per: float
+    output: str
+    ssh_stdout: StringIO = StringIO()
```
cderb, line 491: ssh_stdout: StringIO
is sufficient

cderb (approved):
Technically correct now. Would request reducing commented verbose lines.

PR #887, opened 2023-04-21
Sp/tuna :machine management interfecae - Type hint

technicalgrp89:
machine_management_interface.py  
- code base is supoorted for type hints
- mypy compatible


technicalgrp89:
code coverage % test fails. Opened the PR for code review purpose.

alexandraBara:
> code coverage % test fails. Opened the PR for code review purpose.

@technicalgrp89  please merge with develop where a 0.5% codecov tolerance has been added. This should allow for the codecov to pass here, if not then maybe some code changes need to be addressed

tuna/machine_management_interface.py
```diff
@@ -217,37 +253,36 @@ def run_ipmi_command(self, command):
                 GATEWAY_IP, GATEWAY_PORT, GATEWAY_USER)
 
           ssh = MachineManagementInterface.gateway_session
-          _, out_ch, err_ch = ssh.exec_command(cmd, timeout=SSH_TIMEOUT)
-          error = [x.strip() for x in err_ch.readlines()]
-          error = '\n'.join(error)
-          exit_status = out_ch.channel.exit_status
+          if ssh is not None:
+            _, o_var, e_var = ssh.exec_command(cmd, timeout=SSH_TIMEOUT)
```
alexandraBara: why rename out_ch and err_ch, they are more readable that way compared to o_var and e_varl. Please revert renaming.

tuna/machine_management_interface.py
```diff
@@ -217,37 +253,36 @@ def run_ipmi_command(self, command):
...
+      ret_ipmi = self.run_ipmi_command("chassis power cycle")
+      ret_status = True
     else:
-      ret = self.run_bmc_command("chassisstate")
-    return ret
+      ret_bmc = self.run_bmc_command("chassisoff")
+      ret_bmc |= self.run_bmc_command("chassison")
+    if ret_status:
+      return ret_ipmi
+    return ret_bmc
```
alexandraBara: the previous code only returned one variable: ret, you seem to be returning at least ret_ipmi or ret_bmc. Please only use one variable to store the return code.

alexandraBara (changes requested):

tuna/machine_management_interface.py
technicalgrp89: pylint std rules, does not allow us to use single name for various function returns. I had rename

tuna/machine_management_interface.py
technicalgrp89: This is also, for the same reason. same name is used in multiple places , with different function calls. To address the same, variable renaming is needed

tuna/machine_management_interface.py
alexandraBara: I have tested it myself, and it does work. Please see code below:
```
 def restart_server(self) -> int:
    """ Method to restart a remote machine """
    #ret_ipmi: int
    #ret_bmc: int
    ret: int
    #ret_status: bool = False

    if self.backend == MgmtBackend.IPMI:
      ret = self.run_ipmi_command("chassis power cycle")
      #ret_status = True
    else:
      ret = self.run_bmc_command("chassisoff")
      ret |= self.run_bmc_command("chassison")
    #if ret_status:
    #  return ret_ipmi
    return ret

```

technicalgrp89: Yes, It do. If we look at code readability persepective,  as the functions called are different having same name may confuse. To avoid this, I renamed them depending on the function call.

JehandadKhan: @technicalgrp89 Renaming variables due to the function being called does not make sense to. Since the variable is essentially returning the status of the current function, there should be only one variable. Moreover, this reduces the lines of code considerably, therefore I think what Alex is suggesting makes sense here.

technicalgrp89: ret value changed to Optional[int]

alexandraBara (approved):

iq136boy (approved):

reidkwja (approved):

PR #894, opened 2023-05-02
Integrating export_db subcmd to run through MIOpen 

reidkwja:
./tuna/miopen/subcmd/export_db.py
to
./tuna/go_fish.py miopen export_db


tuna/miopen/miopen_lib.py
```diff
@@ -117,6 +118,9 @@ def parse_args(self):
     subcommands.add_subcommand('load_job',
                                get_load_job_parser(),
                                required=False)
+    subcommands.add_subcommand('export_db',
+                               get_export_db_parser(),
+                               required=False)
```
alexandraBara, line 124: in the run() function you need to call :
```
if self.args.subcommand is not None and self.args.subcommand == 'export_db':
      run_export_db([...])
      return None
```
which would be similar to run_import_configs and run_load_job!!! This will trigger the library to call your script code and not try and launch a worker.

alexandraBara (changes requested):

tests/test_export_db.py
```diff
@@ -0,0 +1,229 @@
...
+    build_miopen_kdb, write_kdb, export_kdb, create_sqlite_tables, get_cfg_dict)
+from tuna.utils.db_utility import DB_Type
+from tuna.miopen.db.tables import MIOpenDBTables, ConfigType
+from tuna.dbBase.sql_alchemy import DbSession
+from tuna.miopen.db.find_db import ConvolutionFindDB
+from utils import add_test_session, DummyArgs
+
+
+@pytest.fixture
+def mock_args():
```
alexandraBara: @reidkwja  we already have something like this in /test/utils.py. If these are used for go_fish you can use the GoFishArgs

tests/test_export_db.py
```diff
@@ -0,0 +1,229 @@
...
+  actual_filename = get_filename(arch, num_cu, filename, ocl, db_type)
+
+  assert actual_filename == expeted_filename, f"expected {expeted_filename}, but got {actual_filename}"
+
+
+with DbSession() as session:
+  session.add(fdb_entry)
+  session.commit()
+
+with DbSession() as session:
```
alexandraBara: instead of nesting the test calls inside the with statement, please create a session and pass it around to the tests. You can then do session.close or something like that

tests/test_export_db.py
```diff
@@ -0,0 +1,229 @@
...
+    alg_lists = get_fdb_alg_lists(fdb_query, logger)
+    miopen_fdb = build_miopen_fdb(alg_lists, logger)
+    fdb_file = write_fdb(mock_args.arch, mock_args.num_cu, mock_args.opencl,
+                         miopen_fdb, mock_args.filename)
+    fdb_exported = export_fdb(dbt, mock_args, logger)
+    assert miopen_fdb is not None, f"failed to build miopen_fdb, Got {type(miopen_fdb)}"
+    assert fdb_file is not None
+    assert fdb_exported is not None
+
+  def test_buid_miopen_kdp(mock_args):
```
alexandraBara:
```suggestion
  def test_buid_miopen_kdb(mock_args):
```

alexandraBara (changes requested):

tuna/miopen/yaml_files/sample1.yaml
```diff
@@ -33,6 +33,10 @@ load_job:
   fin_steps: false
   tag: 'someTag'
 
+export_db:
+  enabled: true
+  file_name: 'fdb_nhwc'
```
alexandraBara: where is this file? what is the extension?

alexandraBara (changes requested):
Is this linted? is there a pylint call for this file in vars/groovy?

tests/test_export_db.py
```diff
@@ -0,0 +1,192 @@
...
+  ocl = True
+  db_type = DB_Type.FIND_DB
+
+  expeted_filename = "tuna_1.0.0/gfx900_64.OpenCL.fdb.txt"
+  actual_filename = get_filename(arch, num_cu, filename, ocl, db_type)
+
+  assert actual_filename == expeted_filename, f"expected {expeted_filename}, but got {actual_filename}"
+
+
+with DbSession() as session:
```
alexandraBara: please create a session and pass it around to the test cases, dont include the test cases in the with statement

tuna/miopen/yaml_files/sample1.yaml
reidkwja: @alexandraBara 
Using this args:
` "./tuna/go_fish.py miopen export_db -a ${arch} -n ${num_cu} -f --session_id ${sesh1} --filename fdb_nhwc"`

tests/utils.py
```diff
@@ -155,3 +155,54 @@ def add_test_session(arch='gfx908', num_cu=120, label=None):
...
+        'pad_w': 0,
+        'in_h': 8,
+        'conv_stride_h': 1,
+        'fil_h': 3,
+        'pad_h': 0,
+        'spatial_dim': 3,
+        'in_d': 8,
+        'conv_stride_d': 1,
+        'fil_d': 3,
+        'pad_d': 0
```
alexandraBara: this is counterintuitive, declare these as class vars(like you do with valid) and then return the dict through using vars(self). We dont want to hardcode it quite this hard :)

tests/utils.py
```diff
@@ -155,3 +155,54 @@ def add_test_session(arch='gfx908', num_cu=120, label=None):
...
+    }
+
+
+class TensorEntry:
+  id = 1
+
+  @staticmethod
+  def to_dict(ommit_valid=False):
```
alexandraBara: same here

alexandraBara (changes requested):

tuna/miopen/yaml_files/sample1.yaml
alexandraBara: doesnt it need an extension though? what kind of file is it? .txt? .db?

alexandraBara (approved):
@cderb can you please review the unit tests that were added

tests/test_export_db.py
```diff
@@ -0,0 +1,188 @@
...
+  assert miopen_fdb is not None, f"failed to build miopen_fdb, Got {type(miopen_fdb)}"
+  assert fdb_file is not None
+  assert fdb_exported is not None
+
+
+def test_buid_miopen_kdb():
+  fdb_query = get_fdb_query(dbt, args, logger)
+  alg_lists = get_fdb_alg_lists(fdb_query, logger)
+  build_mioopen_kdp = build_miopen_kdb(dbt, alg_lists, logger)
+  assert build_mioopen_kdp is not None
```
cderb: Could use some arbitrary data tests.

tests/test_export_db.py
```diff
@@ -0,0 +1,188 @@
...
+def test_build_export_miopen_fdp():
+  fdb_query = get_fdb_query(dbt, args, logger)
+  alg_lists = get_fdb_alg_lists(fdb_query, logger)
+  miopen_fdb = build_miopen_fdb(alg_lists, logger)
+  fdb_file = write_fdb(args.arch, args.num_cu, args.opencl, miopen_fdb,
+                       args.filename)
+  fdb_exported = export_fdb(dbt, args, logger)
+  assert miopen_fdb is not None, f"failed to build miopen_fdb, Got {type(miopen_fdb)}"
+  assert fdb_file is not None
+  assert fdb_exported is not None
```
cderb: Could use specific data cases.

cderb (approved):
Will need to be resolved with 890, which removes get_fdb_alg_lists and rewrites build_miopen_fdb.
Some test cases with data also exist in 890.

cderb (approved):

PR #896, opened 2023-05-09
Sp/session mixin typehint support

technicalgrp89:
type annotations support
mypy compatible code
utils.groovy - mypy cmd added


tuna/session_mixin.py
```diff
@@ -82,7 +88,7 @@ def insert_session(self):
         LOGGER.warning("Err occurred trying to add new session: %s \n %s", err,
                        self)
         session.rollback()
-        entry = self.get_query(session, type(self), self).one()
+        entry = self.get_query(session, type(self), self).one()  #type: ignore
```
alexandraBara: why is this ignored? this should be of type SessionExample or Session.

tuna/session_mixin.py
```diff
@@ -25,29 +25,35 @@
...
 from tuna.utils.logger import setup_logger
 
-LOGGER = setup_logger('session')
+LOGGER: logging.Logger = setup_logger('session')
 
 
 class SessionMixin():
   """Session Mixin to provide interface for the session table"""
   #pylint: disable=too-few-public-methods
+  #pylint: disable=too-many-instance-attributes
```
alexandraBara: is this needed? Have you tried removing it? doesnt seem like that many instance attributes

alexandraBara (changes requested):

tuna/session_mixin.py
technicalgrp89: removed

tuna/session_mixin.py
technicalgrp89: overload function defiition added

alexandraBara (approved):

iq136boy (approved):

reidkwja (approved):

requirements.txt
```diff
@@ -38,7 +38,7 @@ py==1.10.0
 pyasn1==0.4.4
 pycparser==2.19
 Pygments==2.14.0
-pylint>=2.14.5
+pylint<=2.17.0-dev0,>=2.15.4
```
cderb, line 41: I'm guessing this is the range returned by the error log. Should we be using the range or just >=2.15.4?

cderb (approved):

PR #898, opened 2023-05-11
tuna/tables_interface.py - type hint support

technicalgrp89:
mypy compatible code
type hints support

### DIFF
--- a/tuna/tables_interface.py
+++ b/tuna/tables_interface.py
@@ -26,32 +26,35 @@
 ###############################################################################
 """Module that encapsulates the DB representation for a library"""
 
+from typing import Dict, Set, Any, Optional
+from sqlalchemy.orm.query import Query
 from tuna.dbBase.sql_alchemy import DbSession
 
 
-#pylint: disable=too-few-public-methods
 class DBTablesInterface():
   """Represents db tables interface class"""
 
-  def __init__(self, **kwargs):
+  def __init__(self, **kwargs: Dict[str, Dict[str, Any]]):
     """Constructor"""
     super().__init__()
-    allowed_keys = set(['session_id'])
+    allowed_keys: Set = set(['session_id'])
     self.__dict__.update((key, None) for key in allowed_keys)
 
-    #for pylint
-    self.job_table = None
-    self.session_id = None
-    self.session = None
+    self.job_table: Optional[Dict[str, Any]] = None
+    self.session_id: Optional[int] = None
+    self.session: Query = None
 
     self.__dict__.update(
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
 
-  def set_tables(self, sess_class):
+  def do_nothing(self):
+    """dummy function to avoid R0903"""
+
+  def set_tables(self, sess_class) -> bool:
     """Set appropriate tables based on requirements"""
     if self.session_id is not None:
       with DbSession() as session:
-        query = session.query(sess_class).filter(
+        query: Query = session.query(sess_class).filter(
             sess_class.id == self.session_id)
         self.session = query.one()
     return True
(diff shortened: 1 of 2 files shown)


tuna/tables_interface.py
```diff
@@ -26,32 +26,35 @@
...
+    self.job_table: Optional[Dict[str, Any]] = None
+    self.session_id: Optional[int] = None
+    self.session: Query = None
 
     self.__dict__.update(
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
 
-  def set_tables(self, sess_class):
+  def do_nothing(self):
+    """dummy function to avoid R0903"""
```
alexandraBara: remove and add the pylint ignore if needed

tuna/tables_interface.py
```diff
@@ -26,32 +26,33 @@
...
     super().__init__()
-    allowed_keys = set(['session_id'])
+    allowed_keys: Set = set(['session_id'])
     self.__dict__.update((key, None) for key in allowed_keys)
 
-    #for pylint
-    self.job_table = None
-    self.session_id = None
-    self.session = None
+    self.job_table: Optional[Dict[str, Any]] = None
```
alexandraBara: this is not a dict, this should be an option between: ConvolutionJob, BNJob and FusionJob class defined in /miopen/db/miopen_tables.py. Basically any of the classes that implement the JobMixin Interface

tuna/tables_interface.py
```diff
@@ -26,32 +26,33 @@
...
+    self.session_id: Optional[int] = None
+    self.session: Query = None
```
alexandraBara: this is not a Query its any of the classes that implements the SessionMixin interface, i think they might be ExampleSession and Session from MIOpen or something like that, you would have to look it up.

alexandraBara (changes requested):

tuna/tables_interface.py
```diff
@@ -26,28 +26,33 @@
...
     super().__init__()
-    allowed_keys = set(['session_id'])
+    allowed_keys: Set = set(['session_id'])
     self.__dict__.update((key, None) for key in allowed_keys)
 
-    #for pylint
-    self.job_table = None
-    self.session_id = None
-    self.session = None
+    self.job_table: Optional[Union[ConvolutionJob, BNJob, FusionJob,
```
cderb: Simplify this to the parent class. These are all JobMixin.

tuna/tables_interface.py
```diff
@@ -26,28 +26,33 @@
...
+                                   Job]] = None
+    self.session_id: Optional[int] = None
+    self.session: Optional[Union[Session, SessionExample]] = None
```
cderb: Use base class SessionMixin.

cderb (changes requested):

tuna/tables_interface.py
```diff
@@ -26,28 +26,30 @@
...
+from tuna.session_mixin import SessionMixin
+from tuna.miopen.db.miopen_tables import JobMixin
 
 
 #pylint: disable=too-few-public-methods
 class DBTablesInterface():
   """Represents db tables interface class"""
 
-  def __init__(self, **kwargs):
+  def __init__(self, **kwargs: Dict[str, Dict[str, Any]]):
```
cderb, line 39: @alexandraBara do you know the structure of kwargs here?

cderb (approved):

tuna/tables_interface.py
alexandraBara: looks like currently its just a dict that takes in session_id as a key. I think a dict of str to Any should be comprehensive for future

tuna/tables_interface.py
alexandraBara: this is nice, didnt realize you can do this, simplifies things

alexandraBara (approved):

JehandadKhan (approved):

PR #899, opened 2023-05-16
layout length checking for config import

cderb:
Adds a check in the config driver which will raise an error when the layout string for a config is not the proper length indicated by its spatial_dim.


JehandadKhan (approved):

reidkwja (approved):

alexandraBara (approved):
maybe for the future we could use a map for spatial_dim size to out_layout length, if this ends up being more than 2. Or if its always (2x-1) then we could use that too.

PR #905, opened 2023-05-30
load_job fix

cderb:
Fix to load_job for fin_steps arg.

### DIFF
--- a/tests/test_worker.py
+++ b/tests/test_worker.py
@@ -59,7 +59,7 @@ def add_job(w):
   logger = setup_logger('test_add_job')
 
   dbt = MIOpenDBTables(session_id=w.session_id, config_type=args.config_type)
-  counts = import_cfgs(args, dbt, setup_logger('test_worker'))
+  _ = import_cfgs(args, dbt, setup_logger('test_worker'))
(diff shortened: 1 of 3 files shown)


alexandraBara (approved):

msaudulhassan:
On https://github.com/ROCmSoftwarePlatform/MITuna/blob/c451c85342b2a2743d9c97bb3796f5ce38f8f02a/tuna/miopen/subcmd/load_job.py#L51, should the check be `if args.fin_steps == 'not_fin'` ?

cderb:
> On
> 
> https://github.com/ROCmSoftwarePlatform/MITuna/blob/c451c85342b2a2743d9c97bb3796f5ce38f8f02a/tuna/miopen/subcmd/load_job.py#L51
> 
> , should the check be `if args.fin_steps == 'not_fin'` ?

not_fin is a valid value for the fin_steps argument. This check is correct, it requires the fin_step input not to be an empty string. Though notable not_fin is now the default in the input parsing.

msaudulhassan (approved):

reidkwja (approved):

PR #911, opened 2023-06-08
tuna/example/example_lib.py - typehints 

technicalgrp89:
Add Type hints to exsisting code
remove unused code
test against the code changes mypy std's
pass pylint checks
pass ypaf std;s
Pass the CI tests

### DIFF
--- a/tuna/example/example_lib.py
+++ b/tuna/example/example_lib.py
@@ -33,6 +33,8 @@ from typing import Dict, Any, List, Optional, Union
 from tuna.mituna_interface import MITunaInterface
 from tuna.parse_args import TunaArgs, setup_arg_parser, args_check
 from tuna.utils.miopen_utility import load_machines
+from tuna.machine import Machine
+
 from tuna.libraries import Library
 from tuna.utils.db_utility import create_tables
 from tuna.example.example_tables import get_tables
@@ -142,7 +144,7 @@ class Example(MITunaInterface):
     if self.args.add_tables:
       self.add_tables()
       return None
-    machines: int = load_machines(self.args)
+    machines: List[Machine] = load_machines(self.args)
     res = self.compose_worker_list(machines)
     return res
 
--- a/tuna/example/example_lib.py
+++ b/tuna/example/example_lib.py
@@ -27,7 +27,9 @@
 """Example library, integrated with MITuna, runs 'rocminfo' cmd"""
 
 import sys
+import argparse
 
+from typing import Dict, Any, List, Optional
 from tuna.mituna_interface import MITunaInterface
 from tuna.parse_args import TunaArgs, setup_arg_parser, args_check
 from tuna.utils.miopen_utility import load_machines
@@ -45,15 +47,16 @@ class Example(MITunaInterface):
     super().__init__(library=Library.EXAMPLE)
     self.args = None
 
-  def parse_args(self):
+  def parse_args(self) -> None:
     # pylint: disable=too-many-statements
     """Function to parse arguments"""
+    parser: argparse.Namespace
     parser = setup_arg_parser('Example library integrated with MITuna', [
         TunaArgs.ARCH, TunaArgs.NUM_CU, TunaArgs.VERSION, TunaArgs.SESSION_ID,
         TunaArgs.MACHINES, TunaArgs.REMOTE_MACHINE, TunaArgs.LABEL,
         TunaArgs.RESTART_MACHINE, TunaArgs.DOCKER_NAME
     ])
-    group = parser.add_mutually_exclusive_group()
+    group: argparse.Namespace = parser.add_mutually_exclusive_group()
     group.add_argument('--add_tables',
                        dest='add_tables',
                        action='store_true',
@@ -79,17 +82,17 @@ class Example(MITunaInterface):
 
     args_check(self.args, parser)
 
-  def launch_worker(self, gpu_idx, f_vals, worker_lst):
+  def launch_worker(self, gpu_idx: int, f_vals: Dict[str, Any], \
+                    worker_lst: List[str]) -> bool:
     """! Function to launch worker
       @param gpu_idx Unique ID of the GPU
       @param f_vals Dict containing runtime information
       @param worker_lst List containing worker instances
-      @param args The command line arguments
-      @retturn ret Boolean value
+      @return ret Boolean value
     """
 
-    kwargs = self.get_kwargs(gpu_idx, f_vals)
-    worker = ExampleWorker(**kwargs)
+    kwargs: Dict[str, Any] = self.get_kwargs(gpu_idx, f_vals)
+    worker: ExampleWorker = ExampleWorker(**kwargs)
     if self.args.init_session:
       SessionExample().add_new_session(self.args, worker)
       return False
@@ -98,13 +101,12 @@ class Example(MITunaInterface):
     worker_lst.append(worker)
     return True
 
-  def compose_worker_list(self, machines):
+  def compose_worker_list(self, machines) -> Optional[List[str]]:
     # pylint: disable=too-many-branches
     """! Helper function to compose worker_list
-      @param res DB query return item containg available machines
-      @param args The command line arguments
+      @param machines containg available machines
     """
-    worker_lst = []
+    worker_lst: List[str] = []
     for machine in machines:
       if self.args.restart_machine:
         machine.restart_server(wait=False)
@@ -112,11 +114,11 @@ class Example(MITunaInterface):
 
       #determine number of processes by compute capacity
       # pylint: disable=duplicate-code
-      worker_ids = super().get_num_procs(machine)
+      worker_ids: range = super().get_num_procs(machine)
       if len(worker_ids) == 0:
         return None
 
-      f_vals = super().get_f_vals(machine, worker_ids)
+      f_vals: Dict[str, Any] = super().get_f_vals(machine, worker_ids)
 
       for gpu_idx in worker_ids:
         self.logger.info('launch mid %u, proc %u', machine.id, gpu_idx)
@@ -125,13 +127,13 @@ class Example(MITunaInterface):
 
     return worker_lst
 
-  def add_tables(self):
-    ret_t = create_tables(get_tables())
+  def add_tables(self) -> bool:
+    ret_t: bool = create_tables(get_tables())
     self.logger.info('DB creation successful: %s', ret_t)
 
     return True
 
-  def run(self):
+  def run(self) -> Optional[List]:
     # pylint: disable=duplicate-code
     """Main function to launch library"""
     res = None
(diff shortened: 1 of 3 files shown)


tuna/example/example_lib.py
```diff
@@ -45,15 +47,16 @@ def __init__(self):
     super().__init__(library=Library.EXAMPLE)
     self.args = None
 
-  def parse_args(self):
+  def parse_args(self) -> None:
     # pylint: disable=too-many-statements
     """Function to parse arguments"""
+    parser: argparse.Namespace
     parser = setup_arg_parser('Example library integrated with MITuna', [
```
msaudulhassan: `parser: argparse.ArgumentParser` ?

tuna/example/example_lib.py
```diff
@@ -45,15 +47,16 @@ def __init__(self):
...
         TunaArgs.ARCH, TunaArgs.NUM_CU, TunaArgs.VERSION, TunaArgs.SESSION_ID,
         TunaArgs.MACHINES, TunaArgs.REMOTE_MACHINE, TunaArgs.LABEL,
         TunaArgs.RESTART_MACHINE, TunaArgs.DOCKER_NAME
     ])
-    group = parser.add_mutually_exclusive_group()
+    group: argparse.Namespace = parser.add_mutually_exclusive_group()
```
msaudulhassan: perhaps `group: argparse._MutuallyExclusiveGroup`? But it appears all the parents of group are private classes:
``` 
>>> inspect.getmro(group.__class__) 
(<class 'argparse._MutuallyExclusiveGroup'>, <class 'argparse._ArgumentGroup'>, <class 'argparse._ActionsContainer'>, <class 'object'>)
```

tuna/example/example_lib.py
```diff
@@ -125,13 +127,13 @@ def compose_worker_list(self, machines):
...
+    ret_t: bool = create_tables(get_tables())
     self.logger.info('DB creation successful: %s', ret_t)
 
     return True
 
-  def run(self):
+  def run(self) -> Optional[List]:
     # pylint: disable=duplicate-code
     """Main function to launch library"""
     res = None
```
msaudulhassan: missing type hint

tuna/example/example_lib.py
```diff
@@ -143,20 +145,20 @@ def run(self):
     res = self.compose_worker_list(machines)
```
msaudulhassan, line 150: type hints missing

msaudulhassan (changes requested):

tuna/example/example_lib.py
```diff
@@ -98,25 +101,24 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
+    worker_lst: List[str] = []
     for machine in machines:
       if self.args.restart_machine:
         machine.restart_server(wait=False)
         continue
 
       #determine number of processes by compute capacity
       # pylint: disable=duplicate-code
-      worker_ids = super().get_num_procs(machine)
+      worker_ids: range = super().get_num_procs(machine)
```
cderb: A range would require a set of numbers separated by a stride.
get_num_procs wouldn't guarantee this, so a list would be more appropriate.

tuna/example/example_lib.py
```diff
@@ -125,38 +128,38 @@ def compose_worker_list(self, machines):
...
     # pylint: disable=duplicate-code
     """Main function to launch library"""
-    res = None
+    res: Union[List[str], None]
     self.parse_args()
     if self.args.add_tables:
       self.add_tables()
       return None
-    machines = load_machines(self.args)
+    machines: int = load_machines(self.args)
```
cderb: load_machines returns List[Machine]

cderb (changes requested):

tuna/example/example_lib.py
```diff
@@ -125,38 +130,38 @@ def compose_worker_list(self, machines):
...
-    envmt = []
+    envmt: List[str] = []
     return envmt
 
-  def get_kwargs(self, gpu_idx, f_vals):
+  def get_kwargs(self, gpu_idx: int, f_vals: Dict[str, Any]) -> Dict[str, Any]:
     """! Helper function to set up kwargs for worker instances
       @param gpu_idx Unique ID of the GPU
       @param f_vals Dict containing runtime information
       @param args The command line arguments
```
msaudulhassan: update docstring (@param args is no longer the case)

tuna/example/example_lib.py
```diff
@@ -125,38 +130,38 @@ def compose_worker_list(self, machines):
...
-    machines = load_machines(self.args)
+    machines: List[Machine] = load_machines(self.args)
     res = self.compose_worker_list(machines)
     return res
 
-  def get_envmt(self):
+  def get_envmt(self) -> List[str]:
     # pylint: disable=unused-argument
     """! Function to construct environment var
        @param args The command line arguments
```
msaudulhassan: same

tuna/example/example_lib.py
```diff
@@ -98,25 +104,24 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
+    worker_lst: List[str] = []
     for machine in machines:
       if self.args.restart_machine:
         machine.restart_server(wait=False)
         continue
 
       #determine number of processes by compute capacity
       # pylint: disable=duplicate-code
-      worker_ids = super().get_num_procs(machine)
+      worker_ids: List = super().get_num_procs(machine)
```
msaudulhassan, line 121: worker_ids appears to be `Union[None, typing.Iterable]`

tuna/example/example_lib.py
```diff
@@ -98,25 +104,24 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
       if len(worker_ids) == 0:
```
msaudulhassan, line 122: While you are at it, check for both `None` and nonemptiness in L121:
```
if not worker_ids:
  return None
```

tuna/example/example_lib.py
```diff
@@ -98,25 +104,24 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
     worker_lst.append(worker)
     return True
 
-  def compose_worker_list(self, machines):
+  def compose_worker_list(self, machines) -> Optional[List[str]]:
```
msaudulhassan: def compose_worker_list(self, machines) -> Optional[List[ExampleWorker]]:

tuna/example/example_lib.py
```diff
@@ -98,25 +104,24 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
     # pylint: disable=too-many-branches
     """! Helper function to compose worker_list
-      @param res DB query return item containg available machines
-      @param args The command line arguments
+      @param machines containg available machines
     """
-    worker_lst = []
+    worker_lst: List[str] = []
```
msaudulhassan: same

msaudulhassan (changes requested):

tuna/example/example_lib.py
technicalgrp89: @cderb changing range from list to above.

tuna/example/example_lib.py
```diff
@@ -98,65 +105,64 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
       if self.args.restart_machine:
         machine.restart_server(wait=False)
         continue
 
       #determine number of processes by compute capacity
       # pylint: disable=duplicate-code
-      worker_ids = super().get_num_procs(machine)
-      if len(worker_ids) == 0:
+      worker_ids: Union[Iterable, None] = super().get_num_procs(machine)
+      if (isinstance(worker_ids, Sequence) and len(worker_ids) == 0 and
```
msaudulhassan: I think 
```
if not worker_ids:
  return None
```
would suffice, but I'd let @cderb comment on this

msaudulhassan (approved):

tuna/example/example_lib.py
msaudulhassan: @technicalgrp89  Come to think of it, your change to the if condition might break the code.

technicalgrp89: When we change worker_id: list to Union[Iterable, None] , leads to error "it is not sized" parameter. to resolve the error, I had to introduce the check of isinstance(worker_ids, Sequence) and len(worker_ids) == 0.

@msaudulhassan how about keeping the worker_id:list. this would not raise any mypy error and CI is also passing

technicalgrp89: > @technicalgrp89 Come to think of it, your change to the if condition might break the code.

CI is passing.

msaudulhassan: > When we change worker_id: list to Union[Iterable, None] , leads to error "it is not sized" parameter. to resolve the error, I had to introduce the check of isinstance(worker_ids, Sequence) and len(worker_ids) == 0.
> 
> @msaudulhassan how about keeping the worker_id:list. this would not raise any mypy error and CI is also passing

image

I believe your mypy error can be fixed by using `typing.Sequence` instead of `typing.Instance`

> > @technicalgrp89 Come to think of it, your change to the if condition might break the code.
> 
> CI is passing.

Double check with Chris, but here's the thing. What if worker_ids is None? Then Line 127 will error out

tuna/example/example_lib.py
```diff
@@ -79,17 +86,17 @@ def parse_args(self):
 
     args_check(self.args, parser)
 
-  def launch_worker(self, gpu_idx, f_vals, worker_lst):
+  def launch_worker(self, gpu_idx: int, f_vals: Dict[str, Any], \
+                    worker_lst: List[str]) -> bool:
```
cderb: worker_list: List[ExampleWorker] ...

tuna/example/example_lib.py
cderb: Here's a compromise:
got to the implementation of get_num_procs
cast the return value to a list
Instead of returning None, return an empty list []

tuna/example/example_lib.py
```diff
@@ -98,65 +105,64 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
+      worker_ids: Union[Iterable, None] = super().get_num_procs(machine)
+      if (isinstance(worker_ids, Sequence) and len(worker_ids) == 0 and
+          not worker_ids):
         return None
 
-      f_vals = super().get_f_vals(machine, worker_ids)
+      f_vals: Dict[str, Any] = super().get_f_vals(machine, worker_ids)
 
-      for gpu_idx in worker_ids:
+      for gpu_idx in worker_ids or []:
```
cderb: revert this ...

tuna/example/example_lib.py
```diff
@@ -98,65 +105,64 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
 
     return True
 
-  def run(self):
+  def run(self) -> Optional[List]:
     # pylint: disable=duplicate-code
-    """Main function to launch library"""
-    res = None
+    """Main run function of example_lib"""
+    res: Union[List[str], None]
```
cderb: Optional[List] ?

cderb (changes requested):

tuna/example/example_lib.py
```diff
@@ -98,25 +104,25 @@ def launch_worker(self, gpu_idx, f_vals, worker_lst):
...
     return True
 
-  def compose_worker_list(self, machines):
+  def compose_worker_list(self, machines) -> Optional[List[ExampleWorker]]:
     # pylint: disable=too-many-branches
     """! Helper function to compose worker_list
-      @param res DB query return item containg available machines
-      @param args The command line arguments
+      @param machines containg available machines
+      @returns list of actual available machine list
```
cderb: Unclear what this is describing.
machines - list of available machine objects
return - list of worker objects

tuna/example/example_lib.py
```diff
@@ -125,38 +131,38 @@ def compose_worker_list(self, machines):
 
     return worker_lst
 
-  def add_tables(self):
-    ret_t = create_tables(get_tables())
+  def add_tables(self) -> bool:
+    """! this is a funtion specific to each lib that uses get_tables,to
```
cderb: spellcheck function
more concisely: Generates the library specific schema to the connected SQL server.

cderb (approved):

PR #913, opened 2023-06-13
Small Tuning Fixes

cderb:
Fix session init pipeline by propagating parameters.
Add dependencies to docker.

### DIFF
--- a/.jenkins/init_session
+++ b/.jenkins/init_session
@@ -23,7 +23,7 @@ def initSession(){
 
   def docker_run_args = "--network host  --dns 8.8.8.8 --device=/dev/kfd --device /dev/dri:/dev/dri:rw --volume /dev/dri:/dev/dri:rw --group-add video -e TUNA_DB_HOSTNAME=${db_host} -e TUNA_DB_NAME=${params.db_name} -e TUNA_DB_USER_NAME=${db_user} -e TUNA_DB_PASSWORD=${db_password} -e gateway_ip=${gateway_ip} -e gateway_port=${gateway_port} -e gateway_user=${gateway_user} -e TUNA_LOGLEVEL=${params.tuna_loglevel}"
 
-  def num_session = utils.runsql("SELECT count(*) from session where reason='${job_label}';")
+  def num_session_prev = utils.runsql("SELECT count(*) from session where reason='${job_label}';")
 
   if(params.arch != '' && params.num_cu != '')
   {
@@ -36,9 +36,9 @@ def initSession(){
     sh "srun --no-kill -p ${slurm_partition} -N 1 -l bash -c 'docker run ${docker_run_args} ${tuna_docker_name} ./tuna/go_fish.py miopen --init_session -l ${job_label}'"
   }
 
-  def num_session2 = utils.runsql("SELECT count(*) from session where reason='${job_label}';")
-  println "Number of ${job_label} sessions: ${num_session2}"
-  if(num_session.toInteger() >= num_session2.toInteger()){
+  def num_session_now = utils.runsql("SELECT count(*) from session where reason='${job_label}';")
+  println "Number of ${job_label} sessions: ${num_session_now}"
+  if(num_session_prev.toInteger() >= num_session_now.toInteger()){
       error("Session is not correctly initialized with the reason:${job_label}")
   }
 }
--- a/.jenkins/init_session
+++ b/.jenkins/init_session
@@ -63,6 +63,12 @@ def VerifyArgs()
 
 pipeline {
   agent { node { label 'mysql' } }
+  environment {
+      db_name = "${params.db_name}"
+      db_host = "${params.db_host}"
+      db_user = "${params.db_user}"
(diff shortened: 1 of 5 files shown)


.jenkins/init_session
```diff
@@ -38,7 +38,7 @@ def initSession(){
 
   def num_session2 = utils.runsql("SELECT count(*) from session where reason='${job_label}';")
   println "Number of ${job_label} sessions: ${num_session2}"
-  if(num_session.toInteger() <= num_session2.toInteger()){
+  if(num_session.toInteger() >= num_session2.toInteger()){
```
msaudulhassan: Looks good! Just maybe rename `num_session1` and `num_session2` to some more descriptive?

msaudulhassan (approved):

tuna/miopen/subcmd/update_golden.py
```diff
@@ -244,6 +244,8 @@ def actuator2(func):
       session_retry(session, gold_session_update, functools.partial(actuator2),
                     logger)
 
+  LOGGER.info('Finished Updating conv_golden %s', args.golden_v)
```
reidkwja: LOGGER should be lower case letters aligned with function args
Something like this:
`logger.info('Finished Updating conv_golden %s', args.golden_v)`

reidkwja (changes requested):

Dockerfile
```diff
@@ -179,11 +180,5 @@ ENV PYTHONPATH=/tuna
 
 RUN python3 setup.py install
 
-# Install SolverAnalytics dependencies (this require root access, so,
-# the dependencies must be installed here, and can't be installed in Jenkinsfile)
-RUN git clone https://$FIN_TOKEN:x-oauth-basic@github.com/ROCmSoftwarePlatform/SolverAnalytics.git
-RUN pip3 install --default-timeout=100000 -r SolverAnalytics/requirements.txt
-RUN rm -rf SolverAnalytics
-
```
msaudulhassan, line 187: This can be bad in the long run. If a change to Tuna breaks SolverAnalytics, it might go unnoticed

Dockerfile
```diff
@@ -135,18 +135,19 @@ RUN if [ $BACKEND = "OpenCL" ]; then \
...
+ARG FIN_BRANCH=
+RUN if ! [ -z $FIN_BRANCH ]; then \
+        git pull && git checkout $FIN_BRANCH; \
+    fi
 # Install dependencies
 RUN cmake -P install_deps.cmake 
 
-#Build Fin
 WORKDIR $FIN_DIR/_hip
 RUN CXX=/opt/rocm/llvm/bin/clang++ cmake -DCMAKE_BUILD_TYPE=Debug -DCMAKE_PREFIX_PATH=$MIOPEN_DEPS $FIN_DIR
```
msaudulhassan, line 152: May check if fin exists in $FIN_DIR in the first place? (because if $FIN_BRANCH is empty, fin wouldn't be cloned, and these lines will error out)

msaudulhassan (changes requested):

Dockerfile
cderb: I noticed that the SolverAnalytics repo is mainly used by the test in vars/utils.groovy:323 solversAnalyticsTest
And it is installed at this point anyway.

Dockerfile
cderb: MIOpen now includes Fin as a submodule, so the fin directory should exist going forward. If Fin isn't included by MIOpen then this assumption fails. I think it's ok for the docker to fail when building Fin if it doesn't exist. The alternative would be to fail when checking if the folder is populated. Either way, the non-presence of Fin at this stage necessitates failure. 
This could be made more resilient by defaulting to cloning the Fin repo directly.

We would like to start using the fin included with the MIOpen commit, because it will ensure compatibility. Though I left the option to use a custom Fin commit with FIN_BRANCH.

Dockerfile
msaudulhassan: @cderb Right, but the dependencies for SolverAnalytics are installed here. Installing the deps requires root access and that's why it's done in the DockerFile

cderb: It appears that CI passes without installing the dependencies at this point, Including the solverAnalyticsTest stage. And the docker had to rebuild because I updated the rocm and miopen versions. So this appears to be working fine? Why wouldn't we have root inside the docker container when we are installing SolverAnalytics during the jenkins stage?

msaudulhassan: SolverAnalytics requires numpy and pandas, and it appears MITuna itself now has numpy and pandas included as requirements. That's why the CI is passing. However, if at a later point, SolverAnalytics takes on a new requirement, or Tuna gets rid of numpy/pandas from its requirements.txt, Tuna's CI will start erroring out.

cderb: Let's add the install reqs at the same point in utils.groovy.

msaudulhassan (approved):

reidkwja (approved):

PR #916, opened 2023-07-03
tuna/example/load_job - type definitions added

technicalgrp89:
mypy check compatibility
ypaf,pylint compatible code
mypy command is added in to utils.groovy

### DIFF
--- a/tuna/example/load_job.py
+++ b/tuna/example/load_job.py
@@ -28,18 +28,22 @@
 Script for adding jobs to the MySQL database
 """
 
-from sqlalchemy.exc import IntegrityError
+import logging
+import argparse
 
+from typing import Type
+from sqlalchemy.exc import IntegrityError
 from tuna.utils.logger import setup_logger
 from tuna.parse_args import TunaArgs, setup_arg_parser
 from tuna.utils.db_utility import connect_db
 from tuna.dbBase.sql_alchemy import DbSession
 from tuna.example.tables import ExampleDBTables
+from tuna.example.example_tables import Job
 
-LOGGER = setup_logger('example_load_jobs')
+LOGGER: logging.Logger = setup_logger('example_load_jobs')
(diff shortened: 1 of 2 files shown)


reidkwja (approved):

tuna/example/load_job.py
```diff
@@ -53,19 +56,19 @@ def parse_args():
...
-  args = parser.parse_args()
+  args: argparse.Namespace = parser.parse_args()
   if not args.session_id:
     parser.error('session_id must be specified')
 
   return args
 
 
-def add_jobs(args, dbt):
+def add_jobs(args: argparse.Namespace, dbt: ExampleDBTables) -> int:
```
msaudulhassan: It's seems to be outside the scope of this PR, but passing around `args` doesn't seem ideal. It's unclear what args.* the function needs without going through the code. Had the signature been `def add_jobs(label: str, session_id: int, dbt: XYZ)`, it'd be clear outright what the function needs. Or, perhaps, one can create `TunaNamespace` as a subclass of argparse.Namespace, and pass that around? @cderb

msaudulhassan: `dbt` seems to be `typing.Type[ExampleDBTables]`

msaudulhassan (changes requested):

tuna/example/load_job.py
technicalgrp89: @alexandraBara

cderb: @msaudulhassan fair enough, but should mark that out of scope. Perhaps @technicalgrp89 can add a comment to the function that this update should be made for clarity.

technicalgrp89: I will add the comment

msaudulhassan (approved):

cderb (approved):

PR #917, opened 2023-07-04
Tuna/example/session - Type variables - Type hints added

technicalgrp89:
utils.groovy - mypy command added
tuna/example/session.py - type hints added
mypy,ypaf,pylint compatible code

### DIFF
--- a/tuna/example/session.py
+++ b/tuna/example/session.py
@@ -30,6 +30,7 @@ import argparse
 from typing import Tuple
 from sqlalchemy import UniqueConstraint
 
+from sqlalchemy.orm.session import Session
 from sqlalchemy.orm.query import Query
 from tuna.dbBase.base_class import BASE
 from tuna.session_mixin import SessionMixin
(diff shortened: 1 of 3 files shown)


tuna/example/session.py
```diff
@@ -25,25 +25,30 @@
...
+  __tablename__: str = "session_example"
+  __table_args__: Tuple[UniqueConstraint] = (UniqueConstraint("arch",
+                                                              "num_cu",
+                                                              "rocm_v",
+                                                              "reason",
+                                                              "docker",
+                                                              name="uq_idx"),)
 
-  def get_query(self, sess, sess_obj, entry):
+  def get_query(self, sess: SessionMixin, sess_obj, entry) -> Query:
```
msaudulhassan: missing typehints for sess_obj and entry

msaudulhassan (changes requested):

reidkwja (approved):

tuna/example/session.py
cderb: The session name aliasing is biting here.
sess is a sqlalchemy session
sess_obj is a SessionMixin
I think entry is an arg list

cderb (changes requested):

msaudulhassan (approved):

cderb (approved):

PR #918, opened 2023-07-05
Tuna/example/tables.py - type hints added

technicalgrp89:
mypy command is added to utils.groovy
yapf, mypy, pyliint compatible code

### DIFF
--- a/tuna/example/tables.py
+++ b/tuna/example/tables.py
@@ -25,6 +25,8 @@
 #
 ###############################################################################
 """Module that encapsulates the DB representation"""
+
+from typing import Dict, Any, Type
 from tuna.tables_interface import DBTablesInterface
 from tuna.example.example_tables import Job
 from tuna.example.session import SessionExample
@@ -34,17 +36,17 @@ from tuna.example.session import SessionExample
 class ExampleDBTables(DBTablesInterface):
   """Represents db tables for example lib"""
 
-  def __init__(self, **kwargs):
+  def __init__(self, **kwargs: Dict[str, Any]) -> None:
     """Constructor"""
     super().__init__(**kwargs)
 
     #for pylint
-    self.job_table = None
-    self.session_table = SessionExample
+    self.job_table: Type[Job] = Job
+    self.session_table: Type[SessionExample] = SessionExample
 
     self.set_tables()
 
-  def set_tables(self, sess_class=SessionExample):
+  def set_tables(self, sess_class=SessionExample) -> None:
     """Set appropriate tables based on requirements"""
     super().set_tables(sess_class)
     self.job_table = Job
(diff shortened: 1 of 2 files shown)


tuna/example/tables.py
```diff
@@ -34,17 +36,17 @@
...
-  def __init__(self, **kwargs):
+  def __init__(self, **kwargs: Dict[str, Any]) -> None:
     """Constructor"""
     super().__init__(**kwargs)
 
     #for pylint
-    self.job_table = None
-    self.session_table = SessionExample
+    self.job_table: Job = None
+    self.session_table: SessionExample = SessionExample
```
msaudulhassan: Would `SessionMixin` be a better typehint for session_table? @cderb

tuna/example/tables.py
```diff
@@ -34,17 +36,17 @@
...
 
-  def __init__(self, **kwargs):
+  def __init__(self, **kwargs: Dict[str, Any]) -> None:
     """Constructor"""
     super().__init__(**kwargs)
 
     #for pylint
-    self.job_table = None
-    self.session_table = SessionExample
+    self.job_table: Job = None
```
msaudulhassan: Why not set `self.job_table: Job = Job` right here and remove self.set_tables()?

msaudulhassan (changes requested):

tuna/example/tables.py
msaudulhassan: Btw, since job_table and session_table are _types_ `Job` and `Whatever`, repsecitvely, and not _objects_ of these types, you need to do `self.job: typing.Type[Job] = Job`

msaudulhassan (changes requested):

tuna/example/tables.py
cderb: That would be more general, but in this case it would always be the specialization, so this is fine.

cderb: Personally I find it redundant. Having the constructor on the line is self explanatory to me. Having the name repeated doesn't help in my opinion.

tuna/example/tables.py
cderb: Perhaps this example is overly trivial. I think Alex wrote it this way to illustrate that these tables could change depending on another variable. In miopen these tables switch depending on the config type.

tuna/example/tables.py
msaudulhassan: But there'd be no warning if I, by mistake, end up changing the variable's type:

```
self.session_table = SessionExample(...)
# ...
self.session_table = SubclassOfSessionMixin(...)
# ....
self.session_table = 2 # could be bad and go unnoticed
```

But if we have
``` 
self.session_table: SessionMixin = SessionExample(...)
# ...
self.session_table = SubclassOfSessionMixin(...)
# ....
self.session_table = 2 # mypy or smtg would complain that self.session_table can only be a SessionMixin
```

technicalgrp89: I have defined Type

reidkwja (approved):

msaudulhassan (approved):

JehandadKhan:
@technicalgrp89  Please review the conflicts.

JehandadKhan:
@technicalgrp89  Please review the conflicts.

PR #926, opened 2023-08-24
Export script & fdb key parsing fixes

cderb:
fdb key parsing: Correct ordering for stride dimensions, add spatial_dim.
kdb export: Code branch for combined skew db entries, storing entries for both skews when fdbs are separate. Remove device id from -mcpu kernel argument.
pdb export: Change method for creating config dict for sqlite to use fdb key.

### DIFF
--- a/tuna/miopen/subcmd/export_db.py
+++ b/tuna/miopen/subcmd/export_db.py
@@ -243,8 +243,6 @@ def build_miopen_kdb(dbt: MIOpenDBTables, find_db, logger: logging.Logger):
       query = session.query(dbt.kernel_cache)\
           .filter(dbt.kernel_cache.kernel_group == fastest_slv.kernel_group)\
           .filter(dbt.kernel_cache.valid == 1)
-      #logger.warning("adding fdb_key:%s, config:%s, solver:%s, kernel groups: %s",
-      #fdb_key, fastest_slv.config, fastest_slv.solver, fastest_slv.kernel_group)
       for kinder in query.all():
         num_kdb_blobs += 1
         kern_db.append(kinder)
--- a/tuna/miopen/subcmd/export_db.py
+++ b/tuna/miopen/subcmd/export_db.py
@@ -39,12 +39,9 @@ from tuna.miopen.db.miopen_tables import GoldenMixin
 from tuna.miopen.db.tables import MIOpenDBTables
 from tuna.miopen.utils.metadata import SQLITE_PERF_DB_COLS
 from tuna.utils.db_utility import get_id_solvers, DB_Type
-from tuna.utils.utility import arch2targetid
 from tuna.utils.logger import setup_logger
 from tuna.miopen.utils.analyze_parse_db import get_config_sqlite, insert_solver_sqlite
-from tuna.miopen.utils.analyze_parse_db import mysql_to_sqlite_cfg
-from tuna.miopen.utils.parsing import parse_pdb_key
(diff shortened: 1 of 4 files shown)


tuna/miopen/subcmd/export_db.py
```diff
@@ -256,6 +243,7 @@ def build_miopen_kdb(dbt: MIOpenDBTables, find_db, logger: logging.Logger):
       query = session.query(dbt.kernel_cache)\
           .filter(dbt.kernel_cache.kernel_group == fastest_slv.kernel_group)\
           .filter(dbt.kernel_cache.valid == 1)
+      #logger.warning("adding fdb_key:%s, config:%s, solver:%s, kernel groups: %s", fdb_key, fastest_slv.config, fastest_slv.solver, fastest_slv.kernel_group)
```
msaudulhassan: might want to delete this comment

tuna/miopen/subcmd/export_db.py
```diff
@@ -459,6 +442,10 @@ def run_export_db(args: argparse.Namespace, logger: logging.Logger):
     except ValueError as terr:
       logger.error(terr)
 
+  args.src_table = dbt.find_db_table
+  if args.golden_v is not None:
+    args.src_table = dbt.golden_table
+
```
msaudulhassan, line 456: This usage of args feels a little off. `args` is a argparse.Namespace object, and so when I see `args.src_table`, I assume it's something supplied by the user as a command-line argument (when it's actually being set here in code).

But I realize Tuna is doing this in other places too, so I'll leave it up to you to change it or leave it as is

tuna/miopen/utils/parsing.py
```diff
@@ -107,7 +110,7 @@ def parse_2d(key, group_count):  #pylint: disable=too-many-locals
   fil_h, fil_w = tmp[3].split('x')
   batchsize = tmp[7]
   pad_h, pad_w = tmp[8].split('x')
-  conv_stride_w, conv_stride_h = tmp[9].split('x')
+  conv_stride_h, conv_stride_w = tmp[9].split('x')
   dilation_h, dilation_w = tmp[10].split('x')
```
msaudulhassan, line 114: So things are no longer backward compatible now?

msaudulhassan (approved):
Just some general comments

tuna/miopen/utils/parsing.py
cderb: This parsing script was used historically to import configs. And you can see it reference in the driver class still. This does mean that any configs imported from a fdb would be inaccurate, they would be different from the expected config if h != w.  However the configs in the conv_config table are still valid (if unexpected) and the results produced by them are valid and have valid fdb keys. Configs imported by driver command are unaffected by this.
This was discovered when updating the export pdb functionality. The configs produced using this parsing method did not match those observed by the previous method.

tuna/miopen/subcmd/export_db.py
cderb: I'll remove, it was a helpful debug log, but to noisy to keep in normally.

tuna/miopen/subcmd/export_db.py
cderb: Fair point. Though I think this is the best way to avoid repeating this assignment in every function.

tuna/miopen/utils/parsing.py
msaudulhassan: I see

reidkwja (approved):
lgtm

PR #931, opened 2023-09-25
Generic Import Configs for Tuna's subcmds

reidkwja:
This PR allows future implementations (similar to implemented MIOpen subcmd changes) for sub-args and creates a generic parser for import configs.


tuna/subcmd/parse_subtuna_args.py
```diff
@@ -0,0 +1,123 @@
...
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+#
+###############################################################################
+"""Module to represent MIOpen subcommands parsers"""
+import jsonargparse
+from tuna.parse_args import TunaArgs, setup_arg_parser
+
+
+def get_import_cfg_parser(
```
cderb: I think it might be better to include this in tuna/parse_args.py

tuna/subcmd/parse_subtuna_args.py
```diff
@@ -0,0 +1,123 @@
...
+  group = parser.add_mutually_exclusive_group()
+
+  group.add_argument('--print_models',
+                     dest='print_models',
+                     action='store_true',
+                     help='Print models from table')
+  group.add_argument('--add_benchmark',
+                     dest='add_benchmark',
+                     action='store_true',
+                     help='Insert new benchmark')
```
cderb: These are options for the Model and Benchmark tables, which are distinctly MIOpen, so lets keep them in parse_miopen_args.py
We'll need to split the related functions out of the generic version of import_cfgs.py as well.

tuna/subcmd/parse_subtuna_args.py
```diff
@@ -0,0 +1,123 @@
...
+                      required=False,
+                      help='Specify model batchsize')
+  parser.add_argument(
+      '-c',
+      '--command',
+      type=str,
+      dest='command',
+      default=None,
+      help='Command override: run a different command on the imported configs',
+      choices=[None, 'conv', 'convfp16', 'convbfp16'])
```
cderb: These are convolution config concepts, we should keep them in the miopen file.

tuna/subcmd/parse_subtuna_args.py
```diff
@@ -0,0 +1,123 @@
...
+  parser.add_argument('-d',
+                      '--driver',
+                      dest='driver',
+                      type=str,
+                      default=None,
+                      help='Specify driver cmd')
```
cderb: This input seems to allow the user to input a single line command to be added as a config. In our import_cfgs.py implementation however it appears to only be implemented for add_benchmark. So we should leave this in the miopen file for now.

tuna/subcmd/parse_subtuna_args.py
```diff
@@ -0,0 +1,123 @@
...
+                      default=None,
+                      required=False,
+                      help='Specify framework version')
+  parser.add_argument('-g',
+                      '--gpu_count',
+                      dest='gpu_count',
+                      type=int,
+                      default=None,
+                      required=False,
+                      help='Specify number of gpus the benchmark runs on')
```
cderb: Related to miopen tables.

tuna/subcmd/parse_subtuna_args.py
```diff
@@ -0,0 +1,123 @@
...
+      '--mark_recurrent',
+      dest='mark_recurrent',
+      action="store_true",
+      help='Indicate whether you want the configs to be marked as recurrent')
+  parser.add_argument('--md_version',
+                      dest='md_version',
+                      type=int,
+                      default=None,
+                      required=False,
+                      help='Specify model version')
```
cderb: miopen table

cderb (changes requested):
Good start.

tuna/parse_args.py
```diff
@@ -164,3 +165,67 @@ def args_check(args: argparse.Namespace, parser: argparse.Namespace) -> None:
     parser.error(
         "When setting up a new tunning session the following must be specified: "\
         "label.")
+
+
+@typing.no_type_check
+def get_import_cfg_parser(
```
reidkwja, line 170: @cderb 

mypy complains about this function when placed in the same file:

`
**mypy tuna/parse_args.py --ignore-missing-imports --follow-imports=skip**
tuna/parse_args.py:179: error: Item "None" of "Optional[Namespace]" has no attribute "add_mutually_exclusive_group"
tuna/parse_args.py:186: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
tuna/parse_args.py:191: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
tuna/parse_args.py:197: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
tuna/parse_args.py:202: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
tuna/parse_args.py:208: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
tuna/parse_args.py:213: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
tuna/parse_args.py:220: error: Item "None" of "Optional[Namespace]" has no attribute "add_argument"
Found 8 errors in 1 file (checked 1 source file)
`

alexandraBara: the setup_arg_parser() function returns an optional Namespace, so when when the obj returned is None you need to do a None check before trying to add_argument/add_mutually_exclusive...etc.

sample:
if group not None:
    group.add_mutually_exclusive..
    group.add_argument..

either the above, or you remove the Optional Namespace from the mypy definition of setup_arg_parser. This should be up for debate and we need to investigate why it would ever be None.

tuna/parse_args.py
```diff
@@ -164,3 +165,67 @@ def args_check(args: argparse.Namespace, parser: argparse.Namespace) -> None:
...
+                      type=int,
+                      default=None,
+                      required=False,
+                      help='Specify model batchsize')
+  parser.add_argument('-f',
+                      '--file_name',
+                      type=str,
+                      dest='file_name',
+                      help='File to import')
+  parser.add_argument('--fw_version',
```
alexandraBara: needs to stay with --framework

alexandraBara (changes requested):

tuna/parse_args.py
cderb: We could assert that the parser isn't none and re-cast it to namespace.

tuna/parse_args.py
```diff
@@ -164,3 +165,67 @@ def args_check(args: argparse.Namespace, parser: argparse.Namespace) -> None:
...
+                      '--batches',
+                      type=str,
+                      dest='batches',
+                      help='Batch sizes to iterate over in the given configs')
+  parser.add_argument('--batchsize',
+                      dest='batchsize',
+                      type=int,
+                      default=None,
+                      required=False,
+                      help='Specify model batchsize')
```
cderb: Lets keep batches in the miopen parser.

tuna/parse_args.py
```diff
@@ -164,3 +165,67 @@ def args_check(args: argparse.Namespace, parser: argparse.Namespace) -> None:
...
+  """Return parser for import_configs subcommand"""
+
+  parser = setup_arg_parser(
+      'Import MIOpenDriver commands and MIOpen performance DB entries.',
+      [TunaArgs.VERSION, TunaArgs.CONFIG_TYPE],
+      with_yaml=with_yaml)
+
+  group = parser.add_mutually_exclusive_group()
+
+  group.add_argument('--print_models',
```
cderb: print_models is also miopen specific implementation.

tuna/parse_args.py
```diff
@@ -164,3 +165,67 @@ def args_check(args: argparse.Namespace, parser: argparse.Namespace) -> None:
...
         "label.")
+
+
+@typing.no_type_check
+def get_import_cfg_parser(
+    with_yaml: bool = True) -> jsonargparse.ArgumentParser:
+  """Return parser for import_configs subcommand"""
+
+  parser = setup_arg_parser(
+      'Import MIOpenDriver commands and MIOpen performance DB entries.',
```
cderb: Change info string to generic message like 'Import configurations for module'
This message can be overwritten in get_import_cfg_parser_miopen etc.

cderb (changes requested):

tuna/parse_args.py
reidkwja: Changed setup arg parser intakes args from this:


```
def setup_arg_parser(desc: str,
                     arg_list: List[TunaArgs],
                     parser: argparse.Namespace = None,
                     with_yaml: bool = True) -> Optional[argparse.Namespace]:
```

                     
to this 

```
def setup_arg_parser(desc: str,
                     arg_list: List[TunaArgs],
                     parser: Optional[jsonargparse.ArgumentParser] = None,
                     with_yaml: bool = True) -> jsonargparse.ArgumentParser:
```

Tried asserting to not None, then Pylint complained about branches - likely requiring the setup_arg_parser function to be split into parts.

alexandraBara: removing the Optional from the returned value would definitely require more work, lets go with what @cderb said.

alexandraBara:
After looking at the tuna/mlir implementation and these changes here, I am not sure that there is much value in adding generic parser functions without somehow enforcing their use. I think a better design would be something like a parser object that is then specialized for: import_configs, load_jobs and is defined by your library. I am leaving this thought here so we can discuss more offline.

cderb:
> After looking at the tuna/mlir implementation and these changes here, I am not sure that there is much value in adding generic parser functions without somehow enforcing their use. I think a better design would be something like a parser object that is then specialized for: import_configs, load_jobs and is defined by your library. I am leaving this thought here so we can discuss more offline.

I had a similar idea that the parser object would be like the driver class that exists in the miopen directory. If a new module implements a line parser like the driver in miopen, then the import_configs script could be generically applicable.

tuna/miopen/parse_miopen_args.py
```diff
@@ -71,6 +66,17 @@ def get_import_cfg_parser(
                       default=None,
                       required=False,
                       help='Specify model batchsize')
+  parser.add_argument('-f',
```
cderb, line 69: file_name is a generic input. We expect users to submit a file to be imported.

tuna/miopen/parse_miopen_args.py
```diff
@@ -85,62 +91,29 @@ def get_import_cfg_parser(
...
-                      '--framework',
-                      dest='framework',
-                      choices=[frm.value for frm in FrameworkEnum],
-                      help='Specify framework.')
-  parser.add_argument('--fw_version',
-                      dest='fw_version',
-                      type=int,
-                      default=None,
-                      required=False,
-                      help='Specify framework version')
```
cderb, line 103: Would be best to keep these options together as they are related - framework + version

tuna/miopen/parse_miopen_args.py
```diff
@@ -85,62 +91,29 @@ def get_import_cfg_parser(
...
-                      '--model',
-                      dest='model',
-                      choices=[model.value for model in ModelEnum],
-                      help='Specify model')
   parser.add_argument('--md_version',
                       dest='md_version',
                       type=int,
                       default=None,
                       required=False,
                       help='Specify model version')
```
cderb, line 106: Would be best to keep these options together as they are related - model + version

cderb (changes requested):
Will the genericized import_configs.py be addressed in a follow-up PR?

reidkwja:
> Will the genericized import_configs.py be addressed in a follow-up PR?

@cderb @alexandraBara 
Yes, I've planned it out in a way that will address the generic implementation in multiple sub-tasks to reduce PR complexity. I agree that we perhaps might need to either address the MLIR + other subcm libraries if we decided to proceed forward.

alexandraBara:
> > After looking at the tuna/mlir implementation and these changes here, I am not sure that there is much value in adding generic parser functions without somehow enforcing their use. I think a better design would be something like a parser object that is then specialized for: import_configs, load_jobs and is defined by your library. I am leaving this thought here so we can discuss more offline.
> 
> I had a similar idea that the parser object would be like the driver class that exists in the miopen directory. If a new module implements a line parser like the driver in miopen, then the import_configs script could be generically applicable.

I would agree with such a design, this would allow for a 'generic' import_configs with specialized parsers as @cderb mentioned, in the subclasses. I don't like just breaking up the script and pulling out functions, these functions need to be tied to a class somehow otherwise we cannot enforce their use and every new user could/would have to -  just duplicate the work, as it seems like mlir is doing.

alexandraBara:
> > Will the genericized import_configs.py be addressed in a follow-up PR?
> 
> @cderb @alexandraBara Yes, I've planned it out in a way that will address the generic implementation in multiple sub-tasks to reduce PR complexity. I agree that we perhaps might need to either address the MLIR + other subcm libraries if we decided to proceed forward.

Let's discuss in a meeting, i think we need to backtrack a little.

cderb:
> I would agree with such a design, this would allow for a 'generic' import_configs with specialized parsers as @cderb mentioned, in the subclasses. I don't like just breaking up the script and pulling out functions, these functions need to be tied to a class somehow otherwise we cannot enforce their use and every new user could/would have to - just duplicate the work, as it seems like mlir is doing.

With this change I'd like for modules like mlir to be able to write their 'driver' which could parse a file containing the module's configurations. There are some extra flags for parsing in import_configs.py currently which need to be move to a specialization, but the core of import configs could be adapted as something that is generally useable.

End goal: new modules would only write the 'driver', to perform file parsing for configurations. To accomplish this import_configs.py will also need to have its generic features split from its miopen ones. 

Of note is that the miopen specific features in import_configs.py are actually supporting tables and not the config table itself.

reidkwja:
Using a different approach

PR #932, opened 2023-10-04
Initial implementation of GEMM support.

pcf000:
GEMM support.  Basic idea is to put op-specific functions on op-specific classes, and subclass the interface class so the common stuff doesn't have to care.

Abstraction in results classes.
Initial export-as-tsv method.
Implement the load-factor idea, controlling how many workers per GPU, up and down.
Pick up the more detailed arch from ISA listing, if available;  rocmlir expects it.
Add rocmlir/export_configs.py script.
Adjust for the new num_cu field in tuner output.
Note how to get mlir revision.
Add --append to export_configs.py.
Make some output more like existing tuningRunner.py output.
Tweak the convolution config-string output to match existing output a little better.
Use 'tenacity' library for better retry loop on job errors.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -62,3 +62,4 @@ zipp==3.8.1
 coverage==7.0.5
 mysql-connector-python
 prometheus_flask_exporter
+tenacity
--- a/tests/test_importconfigs_rocmlir.py
+++ b/tests/test_importconfigs_rocmlir.py
@@ -60,7 +60,7 @@ def test_importconfigs_rocmlir():
 
 def test_import_conv():
   rocmlir = RocMLIR()
-  assert (rocmlir.add_tables())
+  assert rocmlir.add_tables()
 
   dbt = RocMLIRDBTables(session_id=None, config_type=ConfigType.convolution)
   logger = setup_logger('test_importconfigs')
@@ -76,7 +76,6 @@ def test_import_conv():
   with DbSession() as session:
     after_cfg_num = session.query(dbt.config_table.id).count()
 
-  print(
-      f"after-import count {after_cfg_num} should equal before-import count {before_cfg_num} plus {counts}"
-  )
-  assert (after_cfg_num - before_cfg_num == counts)
+  assert after_cfg_num - before_cfg_num == counts, \
+    f"after-import count {after_cfg_num} should equal " \
+    f"before-import count {before_cfg_num} plus {counts}"
--- a/tests/test_importconfigs_rocmlir.py
+++ b/tests/test_importconfigs_rocmlir.py
@@ -32,10 +32,11 @@ sys.path.append("tuna")
 
 this_path = os.path.dirname(__file__)
 
+from sqlalchemy import func
 from utils import CfgImportArgs
 from tuna.utils.logger import setup_logger
+from tuna.dbBase.sql_alchemy import DbSession
 from tuna.rocmlir.import_configs import import_cfgs
-from tuna.sql import DbCursor
 from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables
 from tuna.rocmlir.rocmlir_lib import RocMLIR
 from tuna.rocmlir.config_type import ConfigType
@@ -63,19 +64,17 @@ def test_import_conv():
 
   dbt = RocMLIRDBTables(session_id=None, config_type=ConfigType.convolution)
   logger = setup_logger('test_importconfigs')
-  find_configs = "SELECT count(*) FROM rocmlir_conv_config;"
 
-  with DbCursor() as cur:
-    cur.execute(find_configs)
-    res = cur.fetchall()
-    before_cfg_num = res[0][0]
+  with DbSession() as session:
+    before_cfg_num = session.query(dbt.config_table.id).count()
 
   args = CfgImportArgs
   args.file_name = "test-conv-configs"
   counts = import_cfgs(args, dbt, logger)
+  print(f"Imported {counts} configs.")
 
-  with DbCursor() as cur:
-    cur.execute(find_configs)
-    res = cur.fetchall()
-    after_cfg_num = res[0][0]
-    assert (after_cfg_num - before_cfg_num == counts)
+  with DbSession() as session:
+    after_cfg_num = session.query(dbt.config_table.id).count()
+
+  print(f"after-import count {after_cfg_num} should equal before-import count {before_cfg_num} plus {counts}")
+  assert (after_cfg_num - before_cfg_num == counts)
--- a/tests/test_importconfigs_rocmlir.py
+++ b/tests/test_importconfigs_rocmlir.py
@@ -76,5 +76,7 @@ def test_import_conv():
   with DbSession() as session:
     after_cfg_num = session.query(dbt.config_table.id).count()
 
-  print(f"after-import count {after_cfg_num} should equal before-import count {before_cfg_num} plus {counts}")
+  print(
+      f"after-import count {after_cfg_num} should equal before-import count {before_cfg_num} plus {counts}"
+  )
   assert (after_cfg_num - before_cfg_num == counts)
--- a/tests/test_importconfigs_rocmlir.py
+++ b/tests/test_importconfigs_rocmlir.py
@@ -35,7 +35,7 @@ this_path = os.path.dirname(__file__)
 from tuna.utils.logger import setup_logger
 from tuna.rocmlir.import_configs import import_cfgs
 from tuna.sql import DbCursor
-from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables
+from tuna.rocmlir.rocmlir_tables import RocMLIRDBTablesConv
 from tuna.rocmlir.rocmlir_lib import RocMLIR
 from utils import CfgImportArgs
 
@@ -60,7 +60,7 @@ def test_import_conv():
   rocmlir = RocMLIR()
   assert (rocmlir.add_tables())
 
-  dbt = RocMLIRDBTables(session_id=None)
+  dbt = RocMLIRDBTablesConv(session_id=None)
   logger = setup_logger('test_importconfigs')
   find_configs = "SELECT count(*) FROM rocmlir_conv_config;"
 
--- a/tests/test_importconfigs_rocmlir.py
+++ b/tests/test_importconfigs_rocmlir.py
@@ -32,12 +32,13 @@ sys.path.append("tuna")
 
 this_path = os.path.dirname(__file__)
 
+from utils import CfgImportArgs
 from tuna.utils.logger import setup_logger
 from tuna.rocmlir.import_configs import import_cfgs
 from tuna.sql import DbCursor
-from tuna.rocmlir.rocmlir_tables import RocMLIRDBTablesConv
+from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables
 from tuna.rocmlir.rocmlir_lib import RocMLIR
-from utils import CfgImportArgs
+from tuna.rocmlir.config_type import ConfigType
 
 SAMPLE_CONV_CONFIGS = """
 # This section of the file comes from resnet50-miopen-configs
@@ -60,7 +61,7 @@ def test_import_conv():
   rocmlir = RocMLIR()
   assert (rocmlir.add_tables())
 
-  dbt = RocMLIRDBTablesConv(session_id=None)
+  dbt = RocMLIRDBTables(session_id=None, config_type=ConfigType.convolution)
   logger = setup_logger('test_importconfigs')
   find_configs = "SELECT count(*) FROM rocmlir_conv_config;"
 
--- a/tests/test_load_job_rocmlir.py
+++ b/tests/test_load_job_rocmlir.py
@@ -50,7 +50,6 @@ def test_cfg_compose():
   """check the config query function for args tags and cmd intake"""
   clear_tables(ConfigType.convolution)
   test_importconfigs_rocmlir()  # to get the configs in place
-  # +++pf: init a session, too.
   count_configs = "SELECT count(*) FROM rocmlir_conv_config;"
   with DbCursor() as cur:
     cur.execute(count_configs)
@@ -59,7 +58,7 @@ def test_cfg_compose():
 
   args = ExampleArgs()
   args.init_session = True
-  args.label = 'test_rocmlir'
+  args.label = "test_load_job"
   args.load_factor = 1
   args.config_type = ConfigType.convolution
   machine = Machine(hostname="test", local_machine=True)
@@ -67,9 +66,9 @@ def test_cfg_compose():
                          session_id=None,
                          machine=machine,
                          num_procs=Value('i', 0))
-  SessionRocMLIR().add_new_session(args, worker)
+  session_id = SessionRocMLIR().add_new_session(args, worker)
 
-  dbt = RocMLIRDBTables(session_id=1)
-  args = argparse.Namespace(session_id=1)
+  dbt = RocMLIRDBTables(session_id=session_id)
+  args = argparse.Namespace(session_id=session_id)
   job_count = add_jobs(args, dbt)
   assert job_count == config_count
--- a/tests/test_load_job_rocmlir.py
+++ b/tests/test_load_job_rocmlir.py
@@ -38,12 +38,12 @@ this_path = os.path.dirname(__file__)
 from tuna.sql import DbCursor
 from tests.test_importconfigs_rocmlir import test_importconfigs_rocmlir
 from tuna.rocmlir.load_job import add_jobs
-from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables, clear_tables
+from tuna.rocmlir.rocmlir_tables import RocMLIRDBTablesConv, clear_tables
 
 
 def test_cfg_compose():
   """check the config query function for args tags and cmd intake"""
-  clear_tables()
+  clear_tables("convolution")
   test_importconfigs_rocmlir()  # to get the configs in place
   # +++pf: init a session, too.
   count_configs = "SELECT count(*) FROM rocmlir_conv_config;"
@@ -52,7 +52,7 @@ def test_cfg_compose():
     res = cur.fetchall()
     config_count = res[0][0]
 
-  dbt = RocMLIRDBTables(session=None)
+  dbt = RocMLIRDBTablesConv(session=None)
   args = argparse.Namespace(session_id=1)
   job_count = add_jobs(args, dbt)
   assert job_count == config_count
--- a/tests/test_load_job_rocmlir.py
+++ b/tests/test_load_job_rocmlir.py
@@ -37,11 +37,11 @@ this_path = os.path.dirname(__file__)
 
 from utils import ExampleArgs
 from multiprocessing import Value
-from tuna.sql import DbCursor
 from tuna.machine import Machine
+from tuna.dbBase.sql_alchemy import DbSession
 from tests.test_importconfigs_rocmlir import test_importconfigs_rocmlir
 from tuna.rocmlir.load_job import add_jobs
-from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables, clear_tables, SessionRocMLIR
+from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables, clear_tables, SessionRocMLIR, ConvolutionConfig
 from tuna.rocmlir.rocmlir_worker import RocMLIRWorker
 from tuna.rocmlir.config_type import ConfigType
 
@@ -50,18 +50,18 @@ def test_cfg_compose():
   """check the config query function for args tags and cmd intake"""
   clear_tables(ConfigType.convolution)
   test_importconfigs_rocmlir()  # to get the configs in place
-  count_configs = "SELECT count(*) FROM rocmlir_conv_config;"
-  with DbCursor() as cur:
-    cur.execute(count_configs)
-    res = cur.fetchall()
-    config_count = res[0][0]
+  with DbSession() as session:
+    config_count = session.query(ConvolutionConfig.id).count()
 
   args = ExampleArgs()
   args.init_session = True
   args.label = "test_load_job"
   args.load_factor = 1
   args.config_type = ConfigType.convolution
-  machine = Machine(hostname="test", local_machine=True)
+  # Fake up a machine.  CI doesn't give access to GPU, thus no arch info.
+  machine = Machine(hostname="dummy", local_machine=True,
+                    arch='gfx908', arch_full='gfx908',
+                    num_cu=12, avail_gpus=[0])
   worker = RocMLIRWorker(config_type=args.config_type,
                          session_id=None,
                          machine=machine,
@@ -71,4 +71,5 @@ def test_cfg_compose():
   dbt = RocMLIRDBTables(session_id=session_id)
   args = argparse.Namespace(session_id=session_id)
   job_count = add_jobs(args, dbt)
+  print(f"job count ({job_count}) should equal config count ({config_count})")
   assert job_count == config_count
--- a/tests/test_load_job_rocmlir.py
+++ b/tests/test_load_job_rocmlir.py
@@ -74,5 +74,5 @@ def test_cfg_compose():
   dbt = RocMLIRDBTables(session_id=session_id)
   args = argparse.Namespace(session_id=session_id)
   job_count = add_jobs(args, dbt)
-  print(f"job count ({job_count}) should equal config count ({config_count})")
-  assert job_count == config_count
+  assert job_count == config_count, \
+    f"job count ({job_count}) should equal config count ({config_count})"
--- a/tests/test_load_job_rocmlir.py
+++ b/tests/test_load_job_rocmlir.py
@@ -59,9 +59,12 @@ def test_cfg_compose():
   args.load_factor = 1
   args.config_type = ConfigType.convolution
   # Fake up a machine.  CI doesn't give access to GPU, thus no arch info.
-  machine = Machine(hostname="dummy", local_machine=True,
-                    arch='gfx908', arch_full='gfx908',
-                    num_cu=12, avail_gpus=[0])
+  machine = Machine(hostname="dummy",
+                    local_machine=True,
+                    arch='gfx908',
+                    arch_full='gfx908',
+                    num_cu=12,
+                    avail_gpus=[0])
   worker = RocMLIRWorker(config_type=args.config_type,
                          session_id=None,
                          machine=machine,
--- a/tests/test_load_job_rocmlir.py
+++ b/tests/test_load_job_rocmlir.py
@@ -35,15 +35,20 @@ sys.path.append("tuna")
 
 this_path = os.path.dirname(__file__)
 
+from utils import ExampleArgs
+from multiprocessing import Value
 from tuna.sql import DbCursor
+from tuna.machine import Machine
 from tests.test_importconfigs_rocmlir import test_importconfigs_rocmlir
 from tuna.rocmlir.load_job import add_jobs
-from tuna.rocmlir.rocmlir_tables import RocMLIRDBTablesConv, clear_tables
+from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables, clear_tables, SessionRocMLIR
+from tuna.rocmlir.rocmlir_worker import RocMLIRWorker
+from tuna.rocmlir.config_type import ConfigType
 
 
 def test_cfg_compose():
   """check the config query function for args tags and cmd intake"""
-  clear_tables("convolution")
+  clear_tables(ConfigType.convolution)
   test_importconfigs_rocmlir()  # to get the configs in place
   # +++pf: init a session, too.
   count_configs = "SELECT count(*) FROM rocmlir_conv_config;"
@@ -52,7 +57,19 @@ def test_cfg_compose():
     res = cur.fetchall()
     config_count = res[0][0]
 
-  dbt = RocMLIRDBTablesConv(session=None)
+  args = ExampleArgs()
+  args.init_session = True
+  args.label = 'test_rocmlir'
+  args.load_factor = 1
+  args.config_type = ConfigType.convolution
+  machine = Machine(hostname="test", local_machine=True)
+  worker = RocMLIRWorker(config_type=args.config_type,
+                         session_id=None,
+                         machine=machine,
+                         num_procs=Value('i', 0))
+  SessionRocMLIR().add_new_session(args, worker)
+
+  dbt = RocMLIRDBTables(session_id=1)
   args = argparse.Namespace(session_id=1)
   job_count = add_jobs(args, dbt)
   assert job_count == config_count
--- a/tests/test_rocmlir.py
+++ b/tests/test_rocmlir.py
@@ -46,7 +46,6 @@ SAMPLE_CONV_CONFIGS = """
 -F 1 -n 256 -c 1024 -H 14 -W 14 -k 2048 -y 1 -x 1 -p 0 -q 0 -u 2 -v 2 -l 1 -j 1 -m conv -g 1 -t 1
 """
 
-
 def test_rocmlir():
   logger = setup_logger('test_rocmlir')
   dbt = RocMLIRDBTables(session_id=None)
--- a/tests/test_rocmlir.py
+++ b/tests/test_rocmlir.py
@@ -37,7 +37,7 @@ from multiprocessing import Value
 
 from tuna.utils.logger import setup_logger
 from tuna.rocmlir.rocmlir_lib import RocMLIR
-from tuna.utils.miopen_utility import load_machines
+from tuna.machine import Machine
 from tuna.dbBase.sql_alchemy import DbSession
 from tuna.rocmlir.rocmlir_tables import SessionRocMLIR, ConvolutionJob, RocMLIRDBTables, clear_tables
 from tuna.rocmlir.load_job import add_jobs
@@ -71,10 +71,10 @@ def test_rocmlir():
   rocmlir.args.label = 'test_rocmlir'
   rocmlir.args.load_factor = 1
   rocmlir.args.config_type = ConfigType.convolution
-  machines = load_machines(rocmlir.args)
+  machine = Machine(hostname="test", local_machine=True)
   worker = RocMLIRWorker(config_type=rocmlir.args.config_type,
                          session_id=None,
-                         machine=machines[0],
+                         machine=machine,
                          num_procs=Value('i', 0))
   SessionRocMLIR().add_new_session(rocmlir.args, worker)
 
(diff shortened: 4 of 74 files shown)


requirements.txt
```diff
@@ -62,3 +62,4 @@ zipp==3.8.1
 coverage==7.0.5
 mysql-connector-python
 prometheus_flask_exporter
+tenacity
```
pcf000, line 65: Used in rocmlir_worker.py.

tuna/machine.py
```diff
@@ -251,7 +251,7 @@ def parse_agents(self) -> dict:  #pylint: disable=too-many-locals
             sub = stack.pop()
 
         if ':' in decoded_line:
-          cols = decoded_line.split(':')
+          cols = decoded_line.split(':', 1)
```
pcf000, line 256: Full arch string is something like "gfx90a:sramecc+:xnack-" and we only want the first part.

tuna/rocmlir/export_configs.py
```diff
@@ -0,0 +1,63 @@
...
+      help='Specify configuration type',
+      default='convolution',
+      choices=['convolution', 'gemm'],  # +++pf: eventually an Enum
+      type=str)
+  parser.add_argument('--append',
+                      dest='append',
+                      action='store_true',
+                      help='Append to file instead of overwriting')
+  args = parser.parse_args()
+  if args.config_type == "convolution":
```
pcf000: More on this later.
1.  RocMLIRDBTables is subclassed by operation.  I think MIOpen has one class and makes operation-related decisions at runtime.  I was experimenting.
2. The export method lives on the subclass.
3. For now, we're writing the same kind of .tsv file as the existing non-Tuna tuning process.
4. An effect of the subclassing is having a --config_type argument everywhere.

tuna/rocmlir/import_configs.py
```diff
@@ -28,141 +28,27 @@
...
-import itertools
+import sys
 
 from sqlalchemy.exc import IntegrityError
 from tuna.dbBase.sql_alchemy import DbSession
 from tuna.utils.db_utility import connect_db
 from tuna.utils.logger import setup_logger
-from tuna.rocmlir.rocmlir_tables import RocMLIRDBTables
-
-## Adapted from perfRunner.getConvConfigurations.
```
pcf000, line 39: These are moved to rocmlir_tables.py, where they're also attached to conv- and gemm-specific classes.

tuna/rocmlir/load_job.py
```diff
@@ -69,19 +76,21 @@ def add_jobs(args, dbt):
                    .filter(dbt.session_table.valid == 1,
                            dbt.session_table.id == args.session_id)
     reasons = query.all()
+    if not reasons:
```
pcf000, line 72: I could easily mess up the session while debugging, and wanted to catch the broken ones early.

tuna/rocmlir/rocmlir_lib.py
```diff
@@ -122,6 +135,14 @@ def compose_worker_list(self, machines) -> Optional[List[RocMLIRWorker]]:
 
       #determine number of processes by compute capacity
       worker_ids: List = machine.get_avail_gpus()
+      factor = self.args.load_factor
```
pcf000, line 132: For experimenting with more than one worker per GPU, or with using fewer than all of the GPUs.  There's some question how much overhead there is in starting a kernel and whether we can exploit the gaps.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -176,8 +178,7 @@ def __repr__(self) -> str:
       'conv_stride_w': '-v',
       'dilation_h': '-l',
       'dilation_w': '-j',
-      'group_size': '-g',
-      'data_type': '-t',
+      'group_size': '-m conv -g',  # Hack to get "-m conv -g 1"
```
pcf000, line 195: Our original conv configs always have that "-m conv" which has no direct equivalent here.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -188,22 +189,83 @@ def __repr__(self) -> str:
...
+    string = "conv"
+    if self.data_type == 'f16':
+      string += 'fp16'
+    elif self.data_type == 'i8':
+      string += 'int8'
+    elif self.data_type == 'bf16':
+      string += 'bfp16'
+    string += " "
+
+    # In options order for canonicalisation, kind of.
```
pcf000, line 215: Always writing config options in the same order, because the .tsv lookup in the performance runs does a simple string match to find the corresponding perf-config.  The order is fixed by using the list of option-field pairs.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -188,22 +189,83 @@ def __repr__(self) -> str:
...
+    for field, flag in self.options.items():
+      value = getattr(self, field, None)
+      if value is not None and flag is not None:
         string += f"{flag} {value} "
-    string += "-m conv"
+    string += "-t 1"  # Dummy "enable timing" option.
     return string
 
+  def parse_line(self, line):
```
pcf000, line 223: Moved from import_configs.py and into a class because gemm has one too, and it's different.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -188,22 +189,83 @@ def __repr__(self) -> str:
...
+    elif operation == 'convbfp16':
+      self.data_type = 'bf16'
+    self.kernel_repeats = 1
+    for flag, value in options.items():
+      field = fields[flag]
+      if field:
+        setattr(self, field, value)
+
+
+class ResultsMixin():  # pylint: disable=too-many-instance-attributes
```
pcf000, line 344: The bulk of results manipulation is the same for both (all?) operations, so abstract it.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -215,7 +277,6 @@ def session(self):
     """session column"""
     return Column(Integer, ForeignKey("session_rocmlir.id"), nullable=False)
 
-  config = Column(Integer, ForeignKey("rocmlir_conv_config.id"), nullable=False)
```
pcf000, line 218: Moving away from foreign-key config IDs and towards local config strings.  We don't really need the ID as such, and as a foreign key it can prevent deleting a config without deleting results.  We could easily want to flush and reload configs while preserving results for the history.  Another thought is to add a session ID to the config tables.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -229,13 +290,155 @@ def get_query(self, sess, result_obj, session_id):
...
     return query
 
-  # +++pf:  rewrite me for tuningRunner.py output!
   def parse(self, lines):
     """parse logger output line for result data """
     line = lines.splitlines()[-1]
     print(f"line being parsed is '{line}'", file=sys.stderr)
     return line.split('\t')
 
+  def export_as_tsv(self, filename, dbt, append=False):
```
pcf000, line 377: Write the relevant contents of the results table in the form that perfRunner.py expects.  For now, we're just dropping the Tuna process in as a replacement for our existing tuning process;  later, we can modify perfRunner.py to query the database directly.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -229,13 +290,155 @@ def get_query(self, sess, result_obj, session_id):
...
+  __tablename__ = "rocmlir_gemm_job"
+  __table_args__ = (UniqueConstraint('config', 'session', name="uq_idx"),)
+
+  config = Column(Integer,
+                  ForeignKey("rocmlir_gemm_config.id"),
+                  nullable=False,
+                  index=True)
+
+
+class GEMMConfig(BASE):
```
pcf000, line 424: Gemm configs are much simpler than convolution configs.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -264,9 +475,146 @@ def set_tables(self, sess_class=SessionRocMLIR):
...
+        if "-out_datatype" not in line:
+          outDataTypeString = f"-out_datatype {datatype} "
+
+        # Strip to avoid spurious spaces
+        oneConfig = f"{dataTypeString}{outDataTypeString}\
+                      {transAString}{transBString}{line}".strip()
+        if oneConfig not in configs:
+          configs.append(oneConfig)
+
+        # Special case to get both i8_i8 and i8_i32, w/o --data-type or output-type-map.
```
pcf000, line 559: Would generally make i8_i8, unless specified in more detail.  If it's general, add the other case too.

tuna/rocmlir/rocmlir_worker.py
```diff
@@ -47,25 +49,34 @@ class RocMLIRWorker(WorkerInterface):
...
+    self.config_type = kwargs['config_type']
     super().__init__(**kwargs)
-    self.set_db_tables()
+    #    self.set_db_tables()
     self.result_attr = [column.name for column in inspect(self.dbt.results).c]
     self.result_attr.remove("insert_ts")
     self.result_attr.remove("update_ts")
 
+
+# Can either have one of these, or --device below, but no combinations.
```
pcf000, line 60: I had been using --device for a while, to specify the GPU a job is to run on.  A Teams discussion brought up the two environment variables and I thought they might work better or be easier.  Turns out they all behave the same, to the level that I can tell, and they also combine in some way when more than one is present, causing errors.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -256,6 +459,14 @@ def __init__(self, **kwargs):
         (key, value) for key, value in kwargs.items() if key in allowed_keys)
     self.set_tables()
 
+  def set_tables(self, sess_class=SessionRocMLIR):
+    """Set appropriate tables based on requirements"""
+    super().set_tables(sess_class)
+
+
+class RocMLIRDBTablesConv(RocMLIRDBTables):
```
pcf000: I was handling several operation-specific functions and tables, and I just didn't want to make a lot of run-time decisions based on the operation when I could abstract it into a class.  Turns out that now I made more init-time decisions, and I'm not sure that's an advantage.

tuna/rocmlir/rocmlir_worker.py
```diff
@@ -47,25 +49,34 @@ class RocMLIRWorker(WorkerInterface):
...
+      self.dbt = RocMLIRDBTablesConv(session_id=self.session_id)
+    else:
+      self.dbt = RocMLIRDBTablesGEMM(session_id=self.session_id)
 
   def update_result_table(self, session, result_str):
     """update results table with individual result entry"""
     obj = self.dbt.results()
 
-    arch, config, perf_config, tflops = obj.parse(result_str)
+    arch, num_cu, config, perf_config, tflops = obj.parse(result_str)
```
pcf000, line 73: Our tuning loop started reporting numCU, too, so we parse it and save it.

tuna/rocmlir/rocmlir_worker.py
```diff
@@ -113,30 +124,39 @@ def step(self):
...
-            self.process_result(string)
-          os.remove(self.output_filename())
-        except OSError as exc:
-          # Most often FileNotFoundError.
-          self.logger.warning(
-              'Exception occurred while saving results of job %s:  %s',
-              self.job.id, exc)
-          self.set_job_state('errored', result=repr(exc))
+      # Retry three times in the case of unhandled exceptions, logging them.
+      for attempt in Retrying(stop=stop_after_attempt(3),
```
pcf000, line 126: Where this came from:  I was observing that out of 300-odd jobs, it wasn't unusual for one to fail around here because it couldn't find the results file that came from the tuning loop.  When I retried the job, it would work.  So let's add a retry loop.  The "tenacity" library abstracts that in a nice way, hence this Retrying object, which goes three times at random intervals before raising the original exception again.

Turns out the real problem was multiple workers on one host, and our tuning loop verifies the final results by running under rocprof, and rocprof always writes to the same default file names.  So the results would sometimes step on each other and that would kill the file and cause the failure.  The better answer is to avoid the collisions in the first place;  for now, that's by using the process ID in the filename, but eventually I'll pipe stdout and avoid named files altogether.

tuna/rocmlir/rocmlir_worker.py
```diff
@@ -161,7 +185,6 @@ def run_cmd(self):
 
   def get_mlir_v(self) -> str:
     """Interface function to get mlir version info"""
-    #_, mlir_ver, _ = self.exec_docker_cmd("cat /opt/rocm/.info/version")
-    mlir_ver = "mlir_v-not-yet-implemented"
-    self.logger.info('Got mlir version: %s', mlir_ver)
-    return mlir_ver
+    _, mlir_hash, _ = self.exec_docker_cmd("git rev-parse HEAD")
```
pcf000, line 191: Finally getting the commit hash of rocMLIR for the version.  Makes me a little nervous to assume what directory we're in, but that happens a lot in this system.

tuna/session_mixin.py
```diff
@@ -41,7 +41,7 @@
 class SessionMixin():
   """Session Mixin to provide interface for the session table"""
 
-  arch: str = Column(String(length=20), nullable=False, server_default="")
+  arch: str = Column(String(length=30), nullable=False, server_default="")
```
pcf000: Full architecture strings like "gfx90a:sramecc+:xnack-" are longer than 20 characters.

alexandraBara:
@pcf000  this is some great work! As I am reviewing it, i noticed that all the tuna/rocmmlir .py files are missing pylint. We strive to cover any py file under pylint to avoid possible bugs, oversight, code duplication etc.

Please see the pylint lines in vars/utils.groovy for examples on how to run pylint.

pcf000:
> Please see the pylint lines in vars/utils.groovy for examples on how to run pylint.

@alexandraBara , thanks.  If I'm reading it right, rocmlir is included in the first pylint call in runLint.  I also run it by hand before making a pull request, because I read the README and am trying to be a good guest.

tuna/rocmlir/export_configs.py
```diff
@@ -0,0 +1,63 @@
...
+def main():
+  """Import conv-configs file into database rocmlir_conv_config table."""
+  # pylint: disable=duplicate-code
+  parser = setup_arg_parser('Export perf-configs from MySQL db',
+                            [TunaArgs.VERSION, TunaArgs.SESSION_ID])
+  parser.add_argument('-f',
+                      '--file_name',
+                      type=str,
+                      dest='file_name',
+                      help='File to import')
```
msaudulhassan: If this is required, I'd set required=True.

tuna/rocmlir/export_configs.py
msaudulhassan: Could you put the code related to config_type in functions? e.g.

``` 
def add_config_type_arg(parser):
  parser.add("--config_type", ...)
  
def get_dbt(config_type, *args, **kwargs):
  if config_type == "convolution":
    return RocMLIRDBTablesConv(*args, **kwargs)
  ...
```
  
and then use these functions everywhere

tuna/rocmlir/import_configs.py
```diff
@@ -184,8 +70,18 @@ def main():
                       type=str,
                       dest='file_name',
                       help='File to import')
```
msaudulhassan: Set required=True

tuna/rocmlir/import_configs.py
```diff
@@ -28,141 +28,27 @@
...
+  configs = dbt.get_configurations(os.path.expanduser(args.file_name))
+  print(configs, file=sys.stderr)
   for line in configs:
     try:
-      config = parse_line(line, dbt)
-
+      config = dbt.config_table()
+      config.parse_line(line)
       with DbSession() as session:
         try:
```
msaudulhassan: I'm haven't really worked with sql_alchemy, but would it be better to move the `with DbSession() as session:` line to outside the for-loop?

tuna/rocmlir/load_job.py
```diff
@@ -69,19 +76,21 @@ def add_jobs(args, dbt):
...
+      raise ValueError(f"No session matching ID {args.session_id}")
     if len(reasons) > 1:
       raise ValueError(f"More than one session matching ID {args.session_id}")
```
msaudulhassan: Would "No session matching ID {args.session_id} with valid jobs" be a better error message? Because if someone sees the message "No session matching ID 2" but there are entries with session ID 2 in their database, they'd very confused as to what's going on, until they look at your query on line 75.

And similarly for the other error message as well.

tuna/rocmlir/rocmlir_worker.py
```diff
@@ -150,7 +170,11 @@ def run_cmd(self):
...
       config_string = config[0].config_string()
-    cmd = env_str + f" python3 ./bin/tuningRunner.py --operation conv \
+    if self.config_type == 'convolution':
+      special_args = "--operation conv"
+    else:
+      special_args = "--operation gemm"
+    cmd = env_str + f" python3 ./bin/tuningRunner.py -q {special_args} \
                      --config='{config_string}' --mlir-build-dir `pwd` \
                      --output={self.output_filename()} --tflops \
                      --rocmlir_gen_flags='--device={self.gpu_id}'"
```
msaudulhassan, line 183:
```
if not os.path.exists("bin/tuningRunner.py"):
  raise FileNotFoundError("Wrong directory")
```

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -188,22 +189,83 @@ def __repr__(self) -> str:
...
+        '-x': 'fil_w',
+        '-p': 'pad_h',
+        '-q': 'pad_w',
+        '-u': 'conv_stride_h',
+        '-v': 'conv_stride_w',
+        '-l': 'dilation_h',
+        '-j': 'dilation_w',
+        '-g': 'group_size',
+        '-m': None
+    }
```
msaudulhassan, line 259: Maybe define this in the class scope

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -229,13 +290,155 @@ def get_query(self, sess, result_obj, session_id):
...
+    fields = {
+        '-transA': 'transpose_A',
+        '-transB': 'transpose_B',
+        '-g': 'group_size',
+        '-m': 'm',
+        '-n': 'n',
+        '-k': 'k',
+        '-t': 'data_type',
+        '-out_datatype': 'out_data_type'
+    }
```
msaudulhassan, line 495: This too can be moved out

msaudulhassan (changes requested):

tuna/machine.py
cderb: good catch

tuna/machine.py
```diff
@@ -293,9 +293,15 @@ def get_properties(self) -> Tuple[List[Dict], List[Dict[str, Any]]]:
...
+        arch = agent['Name']
+        try:
+          arch = agent['ISA Info']['ISA 1']['Name']
+          arch = arch.replace('amdgcn-amd-amdhsa--', '')
+        except KeyError:
+          pass
         details = {
             'rinfo': agent,
-            'arch': agent['Name'],
+            'arch': arch,
```
cderb: Please add a new field here for arch_targetid to store the full arch string. Altering arch would affect existing implementation.

tuna/rocmlir/export_configs.py
cderb: Yes, miopen only has MIOpenDBTables. Depending on the config_type argument to the constructor here, miopen would switch to either convolution or batchnorm. The tablenames used in class are the same, but they would direct to different tables.
If you can't keep the dbt objects in the same class, Saud's suggestion of a function for these 4 lines might be a good compromise.

tuna/rocmlir/import_configs.py
cderb: Yes, every time the 'with' statement is entered a new session is being created. So moving this outside the loop would reduce the overhead.

tuna/rocmlir/load_job.py
cderb: looks good to me
```
select reason from session where valid=1 and id=<id>;
```

cderb (changes requested):

alexandraBara:
> > Please see the pylint lines in vars/utils.groovy for examples on how to run pylint.
> 
> @alexandraBara , thanks. If I'm reading it right, rocmlir is included in the first pylint call in runLint. I also run it by hand before making a pull request, because I read the README and am trying to be a good guest.

I must have missed that, that should do it

tuna/rocmlir/export_configs.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python3
+###############################################################################
+#
+# MIT License
+#
+# Copyright (c) 2022 Advanced Micro Devices, Inc.
```
alexandraBara:
```suggestion
# Copyright (c) 2023 Advanced Micro Devices, Inc.
```
if you grep for 2022 in tuna/rocmlir there are a couple files with 2022 that need to be changed as well

tuna/rocmlir/export_configs.py
alexandraBara: Our code used to have if-else statements based on config_type everywhere. This is one of the reasons why we now make the decision at runtime. It allows us to abstract the backend, so you don't have to refer to concrete-class.attribute everywhere you can just do abstraction.attribute (What attribute points to is decided at runtime and makes for a lot less duplicate code everywhere). Once your code gets larger it will become a pain.

tuna/rocmlir/import_configs.py
```diff
@@ -184,8 +70,18 @@ def main():
                       type=str,
                       dest='file_name',
                       help='File to import')
+  parser.add_argument(
+      '--config_type',
+      dest='config_type',
+      help='Specify configuration type',
+      default='convolution',
+      choices=['convolution', 'gemm'],  # +++pf: eventually an Enum
```
alexandraBara: should be easy to add an enum right about now, before there are more than 2+ 'eventually's in the code.

tuna/rocmlir/load_job.py
```diff
@@ -69,19 +76,21 @@ def add_jobs(args, dbt):
...
     reason = reasons[0].reason
 
     query = session.query(dbt.config_table.id)\
                    .filter(dbt.config_table.valid == 1)
```
alexandraBara, line 79:
```suggestion
    configs = session.query(dbt.config_table.id)\
                   .filter(dbt.config_table.valid == 1).all()
```
removed the intermid var

tuna/rocmlir/rocmlir_lib.py
```diff
@@ -60,6 +60,19 @@ def parse_args(self) -> None:
...
+        '--config_type',
+        dest='config_type',
+        help='Specify configuration type',
+        default='convolution',
+        choices=['convolution', 'gemm'],  # +++pf: eventually an Enum
+        type=str)
+    parser.add_argument('--load_factor',
+                        dest='load_factor',
+                        help='How many workers per GPU',
+                        default=1,
```
alexandraBara:
```suggestion
                        default=1.0,
```

tuna/rocmlir/rocmlir_lib.py
```diff
@@ -170,4 +192,8 @@ def get_kwargs(self, gpu_idx: int, f_vals: Dict[str, Any]) -> Dict[str, Any]:
     """
     kwargs: Dict[str, Any] = super().get_kwargs(gpu_idx, f_vals)
 
+    if self.args.config_type is None:
+      self.args.config_type = "convolution"  # +++pf: eventually an Enum
```
alexandraBara: The defaults should be set right after parse_args is called.

tuna/rocmlir/rocmlir_worker.py
alexandraBara: @cderb can you investigate this?

tuna/rocmlir/rocmlir_worker.py
```diff
@@ -113,30 +124,39 @@ def step(self):
...
+              with open(self.output_filename(), 'r',
+                        encoding='utf8') as results:
+                # https://stackoverflow.com/questions/49902843/avoid-parameter-binding-when-executing-query-with-sqlalchemy
+                string = results.read().replace(':', r'\:')
+                self.set_job_state('completed', result=string)
+                self.process_result(string)
+              os.remove(self.output_filename())
+    # pylint: disable=broad-exception-caught
+    # Not sure what to expect beyond OSError.
+    except Exception as exc:
```
alexandraBara, line 153: We dont like general exception handling around here :) is there a way this can be split into one or more specific exception handling? It would be easier to debug, for instance os.remove/fail can fail in a certain way, and the function calls inside the open() can also be handled differently.

tuna/rocmlir/rocmlir_worker.py
alexandraBara: Do you have the name of the assumed dir? You could add a check beforehand

alexandraBara (changes requested):

tuna/rocmlir/export_configs.py
pcf000: Looks like miopen doesn't have functions that want to vary by operation;  rather, it has a small number of places that check config_type at runtime.  I didn't want to do that, the differences are too large and I didn't want IFs with large bodies.  I'm moving the op-specific functions into the config classes, which will get me back to one DBTables class.  On one hand, it's a little odd because they aren't table functions, but on the other I've already done it with config_string.

tuna/rocmlir/rocmlir_worker.py
pcf000: I'm considering making options to specify tuna and rocMLIR directories.  They won't have fixed names:  some CI machines use /home/jenkins, some use /var/z1_miciadmin.

tuna/rocmlir/rocmlir_worker.py
pcf000: On one hand, the original situation was FileNotFound, but also shouldn't happen any more.  On the other, I haven't run it enough to catch everything that could happen.  An uncaught exception will kill the worker and leave the job in the "running" state, which bothers me.  On the third hand, it really is too broad.

alexandraBara: Im OK with leaving it to broad exception for now

tuna/rocmlir/export_configs.py
pcf000: Done.

tuna/rocmlir/import_configs.py
pcf000: Done.

tuna/rocmlir/import_configs.py
pcf000: Done.  Additional slight change to make a new config-table object on each iteration, because session.add won't add the same object twice, even if its attributes have changed and even if it has committed it.

tuna/rocmlir/load_job.py
pcf000: I did add the word "valid."  I thought about suggesting a SELECT statement to show the situation, but would that be useful or confusing?

tuna/rocmlir/rocmlir_worker.py
pcf000: Done.

tuna/machine.py
pcf000: Done.  Called it "arch_full."

tuna/rocmlir/export_configs.py
pcf000: Done.  Didn't change the non-rocmlir files with 2022, but I easily could if that's handy.

tuna/rocmlir/import_configs.py
pcf000: Done.  Not quite so easy, thanks to arg parsing and argparse vs jsonargparse, and my decision to record config_type in SessionRocMLIR so later steps wouldn't need to specify it when they already specify the session.

tuna/rocmlir/load_job.py
pcf000: Done.

tuna/rocmlir/rocmlir_lib.py
pcf000: Done.

tuna/rocmlir/rocmlir_lib.py
pcf000: This bit goes away now.

pcf000:
Sorry about the barrage of commits.  A lot of my testing used the script I'll call from our CI, and it necessarily had to clone MITuna, which meant I had to push patches.

I think I've done all the requests, aside from moving a couple of function-scope constants out to class scope.  I may still do that.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -570,24 +583,23 @@ class GEMMResults(BASE, ResultsMixin):  # pylint: disable=too-many-instance-attr
 class RocMLIRDBTables(DBTablesInterface):
   """Represents db tables for rocMLIR lib"""
 
-  def __init__(self, **kwargs):
+  def __init__(self, *, config_type=None, session_id, **kwargs):
```
alexandraBara, line 586: if you'd like you can set the config_type default to convolution instead of None.

tuna/rocmlir/rocmlir_worker.py
cderb: My observation is the same, better to pick one of these methods, and not overlay them.

vars/utils.groovy
```diff
@@ -61,7 +61,7 @@ def getMachine() {
 def addMachine(arch, num_cu, machine_ip, machine_local_ip, username, pwd, port) {
     runsql("TRUNCATE machine;")
 // TODO: this should come from different nodes
-    runsql("INSERT INTO machine(hostname, local_ip, local_port,  avail_gpus, user, password, port, arch, num_cu, available, ipmi_inaccessible) VALUES(\'${machine_ip}\', \'${machine_local_ip}\', 22, \'0,1,2,3\', \'${username}\', \'${pwd}\', ${port}, \'${arch}\', ${num_cu}, TRUE, TRUE)" )
+    runsql("INSERT INTO machine(hostname, local_ip, local_port,  avail_gpus, user, password, port, arch, arch_full, num_cu, available, ipmi_inaccessible) VALUES(\'${machine_ip}\', \'${machine_local_ip}\', 22, \'0,1,2,3\', \'${username}\', \'${pwd}\', ${port}, \'${arch}\', \'${arch}\', ${num_cu}, TRUE, TRUE)" )
```
cderb, line 64:
```suggestion
    runsql("INSERT INTO machine(hostname, local_ip, local_port,  avail_gpus, user, password, port, arch, arch_full, num_cu, available, ipmi_inaccessible) VALUES(\'${machine_ip}\', \'${machine_local_ip}\', 22, \'0,1,2,3\', \'${username}\', \'${pwd}\', ${port}, \'${arch}\', \'${arch_full}\', ${num_cu}, TRUE, TRUE)" )
```

cderb (approved):
LGTM

cderb (approved):

tests/test_importconfigs_rocmlir.py
```diff
@@ -60,21 +62,21 @@ def test_import_conv():
...
 
-  with DbCursor() as cur:
-    cur.execute(find_configs)
-    res = cur.fetchall()
-    after_cfg_num = res[0][0]
-    assert (after_cfg_num - before_cfg_num == counts)
+  print(
+      f"after-import count {after_cfg_num} should equal before-import count {before_cfg_num} plus {counts}"
+  )
+  assert (after_cfg_num - before_cfg_num == counts)
```
msaudulhassan:
```
assert after_cfg_num - before_cfg_num == counts, "after-import count {after_cfg_num} should equal before-import count {before_cfg_num} plus {counts}"
```

tests/test_load_job_rocmlir.py
```diff
@@ -35,24 +35,44 @@
...
+                         session_id=None,
+                         machine=machine,
+                         num_procs=Value('i', 0))
+  session_id = SessionRocMLIR().add_new_session(args, worker)
+
+  dbt = RocMLIRDBTables(session_id=session_id)
+  args = argparse.Namespace(session_id=session_id)
   job_count = add_jobs(args, dbt)
+  print(f"job count ({job_count}) should equal config count ({config_count})")
   assert job_count == config_count
```
msaudulhassan: same

tests/test_rocmlir.py
```diff
@@ -66,38 +69,52 @@ def test_rocmlir():
...
+                                         .filter(ConvolutionJob.state=='completed')
+    res = query.all()
+    print(f"Should be 0 'completed' jobs and there are {len(res)}")
+    assert len(res) == 0
+    # Because rocMLIR is not in path, error
+    query = session.query(ConvolutionJob).filter(ConvolutionJob.session==session_id)\
+                                         .filter(ConvolutionJob.state=='error')
+    res = query.all()
+    print(f"Should be 6 'error' jobs and there are {len(res)}")
+    assert len(res) == 6
```
msaudulhassan: same

tests/test_rocmlir.py
```diff
@@ -66,38 +69,52 @@ def test_rocmlir():
...
+    worker.join()
+
+  # Deliberately did not supply ./bin/tuningRunner.py from rocMLIR, so we
+  # should see six 'errored' jobs.
+  with DbSession() as session:
+    query = session.query(ConvolutionJob).filter(ConvolutionJob.session==session_id)\
+                                         .filter(ConvolutionJob.state=='completed')
+    res = query.all()
+    print(f"Should be 0 'completed' jobs and there are {len(res)}")
+    assert len(res) == 0
```
msaudulhassan: ..

tuna/rocmlir/rocmlir_lib.py
```diff
@@ -122,6 +129,14 @@ def compose_worker_list(self, machines) -> Optional[List[RocMLIRWorker]]:
 
       #determine number of processes by compute capacity
       worker_ids: List = machine.get_avail_gpus()
+      factor = self.args.load_factor
+      if factor > 1:
+        # LF>1 means multiple workers per GPU.
+        worker_ids = worker_ids * round(factor)
+      else:
+        # LF<1 means use that fraction of GPUs.
+        worker_ids = worker_ids[:round(len(worker_ids) * factor)]
```
msaudulhassan, line 138: Say there is only 1 GPU and LF is 0.5. Then worker_ids will become an empty list. Does that mean the GPU won't be used at all and the script would just quit? Is this the desired behavior?

msaudulhassan (approved):
Some superficial comments, but LGTM

tuna/rocmlir/export_configs.py
pcf000: Updated everything in MITuna/tuna.

alexandraBara (changes requested):
I believe the copyrights do not need to be updated. A copyright year should mark the year that file was written in. @junliume  can confirm that we do not need to update 2022 to 2023?

alexandraBara:
@pcf000  I have looked at our internal guidelines on copyrights. The year should only be updated to something like: 2022-2023 if there have been substantial changes to the file. Otherwise the year should be left as is. Please undo copyright changes :)

cderb (changes requested):
Will need to rollback copyright changes.

pcf000:
No problem, reverted.  At least it (a) served as a ping and (b) got CI to pass fin-find-eval.

cderb (approved):

alexandraBara (approved):

alexandraBara:
The CI fails dues to some permission errors on that node, we will work on fixing that

PR #934, opened 2023-10-20
Dockerfile update

cderb:
Reorganize docker install dependencies.


Dockerfile
```diff
@@ -24,56 +24,54 @@ RUN if ! [ -z $OSDB_BKC_VERSION ]; then \
...
+# Install dependencies
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -f -y --allow-unauthenticated \
+    rocm-dev \
+    rocm-device-libs \
+    rocm-opencl \
+    rocm-opencl-dev \
+    rocm-cmake \
+    && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
```
alexandraBara, line 37: what is the reason for moving this up here?

cderb: At the top we construct two images, one is with our specified version of rocm, the other is without installing rocm (assuming the provided image has already installed rocm). The apt command for the rocm makings were moved up to be inside the initial image construction.

alexandraBara (approved):

msaudulhassan (approved):
looks good

PR #936, opened 2023-10-25
tuna/driver.py class 

reidkwja:
This PR creates an abstract class for  tuna/driver.py and inherits it through miopen/driver/base.py


tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+from tuna.miopen.db.miopen_tables import ConvolutionConfig
+
+LOGGER = setup_logger('driver_base')
+
+
+class DriverBase(ABC):
+  """Represents db tables based on ConfigType"""
+
+  @abstractmethod
+  def parse_fdb_key(self, line: str):
```
alexandraBara: fdb key is an MIOpen thing, does not need to be here

tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+    """Overloaded method.Defined in conv&bn driver child class"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def parse_row(self, db_obj: ConvolutionConfig):
+    """Overloaded method.Defined in conv&bn driver child class"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def set_cmd(self, data_type: str):
```
alexandraBara: where is this coming from?

tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+    """Overloaded method.Defined in conv&bn driver child class"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def config_set_defaults(self):
```
alexandraBara: also this? where is this implemented?

tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+    """Overloaded method.Defined in conv&bn driver child class"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def compose_weight_t(self):
```
alexandraBara: i would move this to the miopen base

tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+    """Overloaded method.Defined in conv&br driver child class"""
+    raise NotImplementedError("Not implemented")
+
+  @staticmethod
+  @abstractmethod
+  def test_skip_arg(tok1: str):
```
alexandraBara: where is this implemented? Any function declared abstract here needs to have an implementation in one of the child classes, otherwise there is no point in declaring it here

tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+
+  @staticmethod
+  @abstractmethod
+  def get_check_valid(tok1: str, tok2: Union[str, int]):
+    """Overloaded method.Defined in conv&br driver child class"""
+    raise NotImplementedError("Not implemented")
+
+  @staticmethod
+  @abstractmethod
+  def get_common_cols() -> List[str]:
```
alexandraBara: also needs to move to miopen base

tuna/driver.py
```diff
@@ -0,0 +1,107 @@
...
+    key: str
+    value: Union[int, str]
+    for key, value in vars(self).items():
+      if key == "_cmd":
+        copy_dict["cmd"] = value
+      else:
+        copy_dict[key] = value
+    return copy_dict
+
+  def __eq__(self, other: object) -> bool:
```
alexandraBara: this needs to be abstract here and implemented in the child classes, how else are you going to compare the members of the child classes in the base class if they are not even defined yet? there is no point in having a base class __eq__ function

alexandraBara (changes requested):

tuna/driver.py
reidkwja: In convolution and batch_norm, moving to miopen/base.py

tuna/driver.py
reidkwja: similar to set_cmd, moving to miopen/base

tuna/driver.py
```diff
@@ -0,0 +1,94 @@
...
+
+LOGGER = setup_logger('driver_base')
+
+
+class DriverBase(ABC):
+  """Represents db tables based on ConfigType"""
+
+  @abstractmethod
+  def construct_driver(self, line: str) -> bool:
+    """Takes a MIOpenDriver cmd or PDB key"""
```
cderb: Takes a line description of a configuration.

tuna/driver.py
```diff
@@ -0,0 +1,94 @@
...
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def construct_driver_from_db(self, db_obj: Any) -> bool:
+    """Takes a <>_config row and returns a driver cmd"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def insert_tensor(self, tensor_dict: dict) -> int:
```
cderb: I'm anticipating that tensors won't be generally applicable for most use cases.

tuna/driver.py
```diff
@@ -0,0 +1,94 @@
...
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def get_weight_t_id(self) -> int:
+    """Build 1 row in tensor table based on layout from fds param
+     Details are mapped in metadata LAYOUT"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def parse_driver_line(self, line: str):
```
cderb: This may be the same as construct_driver in the general case.

cderb (changes requested):
get_db_obj, which is defined in driver/convolution.py and driver/batchnorm.py should be made a requirement in this interface as well.

tuna/miopen/driver/base.py
```diff
@@ -39,11 +39,12 @@
 from tuna.miopen.db.miopen_tables import ConvolutionConfig
 from tuna.miopen.utils.metadata import TENSOR_PRECISION
 from tuna.miopen.utils.parsing import parse_line
+from tuna.driver import DriverBase
 
 LOGGER = setup_logger('driver_base')
 
 
-class DriverBase():
+class MIOpenDriver(DriverBase):
```
alexandraBara, line 49: the __init__  here needs to be moved to the DriverBase class, exactly as it is. Then call super() here.

tuna/driver.py
alexandraBara: ditto! move all the tensor stuff to miopen, insert, compose, get_input_t,...all that.

tuna/miopen/driver/base.py
```diff
@@ -39,11 +39,12 @@
...
   """Represents db tables based on ConfigType"""
   tensor_attr: List[str] = [column.name for column in inspect(TensorTable).c]
```
alexandraBara, line 51: move these inside init

tuna/miopen/driver/base.py
```diff
@@ -167,10 +168,10 @@ def insert_tensor(self, tensor_dict: dict) -> int:
         tid.valid = 1
         key = build_dict_val_key(tid)
         #cache the tensor table to avoid queries
-        if not DriverBase.tensor_id_map:
```
alexandraBara, line 174: once you move tensor_id_map inside __init__ they will become object attributes not class attributes. @cderb  is there a reason to keep them as class attr?

tuna/driver.py
alexandraBara: just a heads up that if you rename this, there will be some code changes in other files

alexandraBara (changes requested):

tuna/miopen/driver/base.py
cderb: It would be optimal to keep tensor_id_map as a class static or a global. This reduces the number of queries to produce the map. As it is, it is only re-queried when there is an integrity error (a miss) when accessing the map. Making it a object attribute would mean that this query would happen for each object individually (which would be slower).

alexandraBara: fine with me

tuna/driver.py
```diff
@@ -0,0 +1,69 @@
...
+
+LOGGER = setup_logger('driver_base')
+
+
+class DriverBase(ABC):
+  """Represents db tables based on ConfigType"""
+
+  @abstractmethod
+  def construct_driver_from_db(self, db_obj: Any) -> bool:
+    """Takes a <>_config row and returns a driver cmd"""
```
msaudulhassan: 1. could you be a little more specific about the type for db_obj?
2. It says this function returns a driver cmd, but the type annotation mentions the return type to be bool

tuna/driver.py
```diff
@@ -0,0 +1,69 @@
...
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def compose_fds(self, tok: list, line: str) -> bool:
+    """Compose fds from driver line"""
```
msaudulhassan: what are fds?

tuna/driver.py
```diff
@@ -0,0 +1,69 @@
...
+
+  @abstractmethod
+  def to_dict(self) -> Dict[str, Union[str, int]]:
+    """Return class to dictionary"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def __eq__(self, other: object) -> bool:
+    """Defining equality functionality"""
+    raise NotImplementedError("Not implemented")
```
msaudulhassan: Is there a way to define in the docstring what equality means in this context? The implementation can be left to the subclass

tuna/driver.py
```diff
@@ -0,0 +1,69 @@
...
+
+  @abstractmethod
+  def get_db_obj(self, keep_id: bool = False) -> ConvolutionConfig:
+    """Return the DB representation of this object"""
```
msaudulhassan: Isn't the return type annotation too specific here?

tuna/miopen/driver/base.py
msaudulhassan, line 51: You might want to initialize the superclass first

msaudulhassan (commented):
Just leaving some superficial comments

tuna/driver.py
cderb: This should likely be a return type of BASE, being the db base class.

tuna/driver.py
cderb: This method could be copied wholesale from it's original location. It doesn't need to be abstract either. Equality means: having all the same field values.
https://github.com/ROCmSoftwarePlatform/MITuna/blob/43d0f7d5c49909615a08cf893ec7b37ce010293c/tuna/miopen/driver/base.py#L308-L314

tuna/driver.py
```diff
@@ -0,0 +1,63 @@
...
+    """Takes a <>_config row and returns a driver cmd"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def compose_fds(self, tok: list, line: str) -> bool:
+    """Compose fds from driver line"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def to_dict(self) -> Dict[str, Union[str, int]]:
```
cderb, line 60: I think it would be good to include some base case code here:
```py
  def to_dict(self) -> Dict[str, Union[str, int]]:
    """Return class to dictionary"""
    copy_dict: Dict[str, Union[str, int]] = {}
    key: str
    value: Union[str, int]
    for key, value in vars(self).items():
      copy_dict[key] = value
    return copy_dict
    ```

File: tuna/driver.py (hunk 15 above)
Comment (cderb) [reply tuna/driver.py]: compose_fds is a parsing function for an MIOpenDriver line
https://github.com/ROCmSoftwarePlatform/MITuna/blob/43d0f7d5c49909615a08cf893ec7b37ce010293c/tuna/miopen/driver/base.py#L257-L268

Lets replace this function with something more generic
```py
def parse_cmd(self, line: str):
    """Parse command line input"""

```

tuna/driver.py
cderb: This is the inverse of get_db_obj
https://github.com/ROCmSoftwarePlatform/MITuna/blob/43d0f7d5c49909615a08cf893ec7b37ce010293c/tuna/miopen/driver/base.py#L121-L128
This test should be true
```py
a = DriverBase(input_line)
db_obj_b = a.get_db_obj()
b = DriverBase()
b.construct_driver_from_db(db_obj_b)
a == b  #True
```
My opinion would be to make this a static method that returns a new DriverBase object built using the sqlalchemy input object.
This is a good opportunity to rename the function / update the string annotation.

cderb (changes requested):

tuna/driver.py
reidkwja: Because this is an abstract class, the class and functions are defined as a template for the subclass to implement them

tuna/driver.py
reidkwja: Abstract class, implementation would make it a concreate class

tuna/driver.py
msaudulhassan: A shorter option might use the `copy` module:
```python
def to_dict(self) -> Dict[str, Union[str, int]]:
  return copy.copy(vars(self))
```
Like chris mentioned, this would provide sort of a default behavior. So the subclass can either use this inherited default behavior, or override it. It also provides some expectation as to what the `to_dict` function is supposed to do

tuna/driver.py
alexandraBara: can u replace DriverBase with something like type(self)? then u dont need to use the concrete class

cderb: @reidkwja I think this will be more like an interface. Methods like this one are universal, so it's ok to have them at this level.

tuna/driver.py
```diff
@@ -0,0 +1,65 @@
...
+class DriverBase(ABC):
+  """Represents db tables based on ConfigType"""
+
+  @abstractmethod
+  def construct_driver_from_db(self, db_obj: Any) -> bool:
+    """Takes a <>_config row and returns a driver cmd"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def compose_fds(self, tok: list, line: str) -> bool:
```
cderb: I think we should remove compose_fds, because it is a utility function to parse_driver_line. parse_driver_line is sufficient at the API level.

tuna/driver.py
```diff
@@ -0,0 +1,65 @@
...
+"""Module that encapsulates the DB representation of a Driver cmd"""
+
+from typing import Union, Dict, Any
+from abc import ABC, abstractmethod
+from tuna.miopen.db.miopen_tables import ConvolutionConfig
+
+
+class DriverBase(ABC):
+  """Represents db tables based on ConfigType"""
+
```
cderb, line 36: IMO we should have the \_\_init__ method from the previous DriverBase here (with some minor changes).
https://github.com/ROCmSoftwarePlatform/MITuna/blob/2d62e3dd5c12b71e4782b7db40a1cbb57fa86bcc/tuna/miopen/driver/base.py#L51-L63
This would ensure that classes that extend this one would declare new objects in the same way.
ex ``driver = DriverConvolution(line=cmd)``

cderb (changes requested):

tuna/driver.py
```diff
@@ -0,0 +1,86 @@
...
+  @abstractmethod
+  def construct_driver(self, line: str) -> bool:
+    """Takes a MIOpenDriver cmd or PDB key"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def construct_driver_from_db(self, db_obj: Any) -> bool:
+    """Takes a <>_config row and returns a driver cmd"""
+    raise NotImplementedError("Not implemented")
```
msaudulhassan: Is the typehint `bool` for the return value correct? The docstring says the function "returns a driver cmd"

tuna/miopen/driver/base.py
```diff
@@ -27,62 +27,67 @@
...
 
+  @abstractmethod
   def set_cmd(self, data_type: str):
     """Overloaded method.Defined in conv&bn driver child class"""
     raise NotImplementedError("Not implemented")
 
+  @abstractmethod
   def config_set_defaults(self):
     """Overloaded method.Defined in conv&bn driver child class"""
     raise NotImplementedError("Not implemented")
```
msaudulhassan: In the docstrings for these functions, could you please say what they are supposed to do?

tuna/miopen/driver/base.py
```diff
@@ -103,6 +108,20 @@ def get_common_cols() -> List[str]:
...
+    direction: str
+    fds, _, direction = get_fds_from_cmd(line)
+    setattr(self, 'direction', DIR_MAP[direction])
+    for key in self.to_dict():
+      if key in fds:
+        setattr(self, key, fds[key])
+
+    pattern_3d = '[0-9]x[0-9]x[0-9]'
+    if search(pattern_3d, line):
+      setattr(self, 'spatial_dim', 3)
```
msaudulhassan: What if the line has patterns like `3x256x256`? Would this be more general:
```python
pattern_3d = "[0-9]+x[0-9]+x[0-9]+"
```

msaudulhassan: Also, maybe @cderb  can answer this better: what if the line has all patterns like `1x256x256`. Then the spatial dim should be 2, right?

tuna/driver.py
```diff
@@ -0,0 +1,86 @@
...
+      return
+
+    if db_obj:
+      if not self.construct_driver_from_db(db_obj):
+        raise ValueError(
+            f"Error creating Driver from db obj: '{db_obj.to_dict()}'")
+      return
+
+    raise ValueError(
+        "Error creating Driver. Driver cmd line or db_obj required")
```
alexandraBara:
```suggestion
    if line:
      if not self.construct_driver(line):
        raise ValueError(f"Error creating Driver from line: '{line}'")
    elif db_obj:
      if not self.construct_driver_from_db(db_obj):
        raise ValueError(
            f"Error creating Driver from db obj: '{db_obj.to_dict()}'")
    else:
        raise ValueError(
        "Error creating Driver. Driver cmd line or db_obj required")
```

alexandraBara: im not sure my indents are correct, but i think u get the point:
if
elif
else

instead of adding those empty return statements

tuna/driver.py
alexandraBara: @cderb 's statement still stands, please update the description. If this file is to be located in /tuna/ then it cannot reference MIOpen stuff.

tuna/driver.py
```diff
@@ -0,0 +1,86 @@
...
+
+  @abstractmethod
+  def construct_driver(self, line: str) -> bool:
+    """Takes a MIOpenDriver cmd or PDB key"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def construct_driver_from_db(self, db_obj: Any) -> bool:
+    """Takes a <>_config row and returns a driver cmd"""
```
alexandraBara:
```suggestion
    """Takes a db row of a configuration and returns the string representation"""
```

tuna/driver.py
```diff
@@ -0,0 +1,86 @@
...
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def parse_driver_line(self, line: str):
+    """Parse line and set attributes"""
```
alexandraBara:
```suggestion
    """Parse configuration line and set attributes"""
```

tuna/driver.py
alexandraBara: the above comment still stands, there is no use in putting this here, we will never (famous last words) compare base classes. This needs a 'Not implemented' here and move the concrete function in the child classes

tuna/miopen/driver/base.py
reidkwja: done!

tuna/driver.py
```diff
@@ -0,0 +1,75 @@
...
+  def construct_driver_from_db(self, db_obj: Any) -> bool:
+    """Takes a db row of a configuration and returns the string representation"""
+    raise NotImplementedError("Not implemented")
+
+  @abstractmethod
+  def parse_driver_line(self, line: str):
+    """Parse configuration line and set attributes"""
+    raise NotImplementedError("Not implemented")
+
+  def to_dict(self) -> Dict[str, Union[str, int]]:
```
alexandraBara, line 60: can you please double check that print(driver) still prints correctly. Otherwise we can overwrite the __str__() function like:
```
def __str__(self):
  return self.to_dict()
```

so we can print drivers in a nice fashion ;)

tuna/driver.py
alexandraBara: @reidkwja you need to update the function description with what @cderb put up here^. This issue is still not resolved until that happens

alexandraBara (changes requested):

tuna/driver.py
reidkwja: done

cderb: This function comment still needs to remove miopen reference. Something like "Parse configuration from input line" should go here.

tuna/driver.py
cderb: Let's remove parse_driver_line from the interface as it is essentially a sub-method of construct_driver.

tuna/driver.py
```diff
@@ -0,0 +1,75 @@
...
+    """Return class to dictionary"""
+    copy_dict: Dict[str, Union[str, int]] = {}
+    key: str
+    value: Union[int, str]
+    for key, value in vars(self).items():
+      if key == "_cmd":
+        copy_dict["cmd"] = value
+      else:
+        copy_dict[key] = value
+    return copy_dict
```
cderb: Let's simplify this to returning a copy of vars(self).items()
_cmd exception is hacky and it's miopen specific.

alexandraBara (approved):
Approved pending addition of __str__() overwrite function as per review

tuna/miopen/driver/base.py
cderb: > What if the line has patterns like `3x256x256`? Would this be more general:
> 
> ```python
> pattern_3d = "[0-9]+x[0-9]+x[0-9]+"
> ```

This is correct, this regex needs '+' to accommodate for larger numbers.
If the spatial dim is 2 the fdb key should not include the depth field.

cderb (changes requested):

tuna/miopen/driver/base.py
cderb: Also why did this function move to base.py from convolution.py?

cderb (approved):
LGTM

PR #937, opened 2023-11-08
Driver layout fix

alexandraBara:
Fixes tensor layouts for convolution when not all 3 layouts are specified through the MIOpenDriver. Behavior is now this: 

- explicit check that all 3 convolution tensor layouts match at the end of the constructor. If different layouts will be supported in the future, this will need to be updated.
- when not all 3 layout are specified, the specified layout will overwrite the default hardcoded layout (NCHW).
- when no layout is specified, the default NCHW populates all 3 tensor layouts (this is already existing behavior).

Appends all 3 layouts when composing a Driver obj from a db_row. Previously in_layout/fil_layout where omitted.
When constructing a driver obj from an fdb_key we now check that all 3 layouts match.

To display the final driver cmd post constructor call do:
```
print(driver)
```


tests/test_driver.py
```diff
@@ -134,3 +135,12 @@ def test_driver():
     driver_3_row = DriverBatchNorm(db_obj=row3)
     #compare DriverBN for same driver cmd built from Driver-line, vs built from that Driver-line's DB row
     assert driver3 == driver_3_row
+
+  fdb1 = "64-75-75-3x3-64-75-75-512-1x1-1x1-1x1-0-NHWC-FP16-W="
+  driver4 = DriverConvolution(fdb1)
+  d4_str = driver4.__str__()
+  assert (d4_str["in_layout" == "NHWC"])
+  assert (d4_str["out_layout" == "NHWC"])
+  assert (d4_str["fil_layout" == "NHWC"])
```
msaudulhassan: Is this right?

tuna/miopen/driver/base.py
```diff
@@ -293,6 +299,27 @@ def compose_fds(self, tok: list, line: str) -> bool:
...
+    for key, value in layout_dict.items():
+      value_set.add(value)
+
+    if len(value_set) != 1:
+      raise ValueError(f"Layouts do not match: {x for x in {layouts}}")
+
+    ly_values = [x for x in layout_dict.values()]
+    for layout in layouts:
+      setattr(self, layout, ly_values[0])
+
```
msaudulhassan: I'm not sure if layout_dict is serving any purpose. Is this correct?
```python
for layout in layouts:
  if layout in line:
    value_set.add(getattr(self, layout))

if len(value_set) != 1:
  raise ValueError(...)

for layout in layouts:
  setattr(self, layout, value_set[0])
```

tests/test_driver.py
```diff
@@ -134,3 +135,12 @@ def test_driver():
...
+  driver5 = DriverConvolution(d4_str)
+  assert driver4 == driver5
```
msaudulhassan: Would it help to also do layout checks on driver5?

msaudulhassan (changes requested):

tests/test_driver.py
msaudulhassan: These asserts will always pass. I think what you want to do would require some string parsing

alexandraBara: i just realized my brackets are off, ill update

cderb (approved):
LGTM
double checking: This ensures that in parsing of the MIOpenDriver command will populate all three layouts correctly.
Parsing for fdb-keys currently has correct behavior for layout parsing?

alexandraBara:
> LGTM double checking: This ensures that in parsing of the MIOpenDriver command will populate all three layouts correctly. Parsing for fdb-keys currently has correct behavior for layout parsing?

Both creating the ConvolutionDriver obj from fdb or Driver cmd would not populate all 3 layouts. However this behavior was only recently introduce since adding NHWC configs. I checked our DB and the only entries where tensor layouts and out_layout does not match is where one or more of those is a 3D layout, but all these entries have valid=0.

msaudulhassan (approved):
lgtm

PR #938, opened 2023-11-08
Pipeline fixes

cderb:
add branch_id definitions to pipeline environments
move getDockerName call inside groovy function
rm old Jenkinsfile
update Dockerfile rocm BKC


vars/utils.groovy
```diff
@@ -47,7 +47,7 @@ def buildSchema(){
 def getDockerName(backend)
 {
     def docker_registry = "${headnode}:5000"
-    def tuna_docker_name = "${docker_registry}/ci-tuna:${branch_id}_${backend}_${env.BUILD_ID}"
```
alexandraBara, line 50: we dont want to use the env build id to differentiate the dockers?

cderb: BUILD_ID is baked into branch_id so I removed it here.

alexandraBara (approved):

technicalgrp89 (approved):

PR #939, opened 2023-11-11
Dockerfile buildspeed updates

cderb:
Use miopen ci docker as base image - removes need to build miopen dependencies + rocm
Option to build a specific rocm maintained, but disabled by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 #default image to ubuntu + install rocm
 ARG BASEIMAGE=rocm/miopen:ci_5450cc
-#ARG ROCM_PRE=0
 
 #FROM ubuntu:20.04 as dtuna-ver-0
 FROM $BASEIMAGE as dtuna-ver-0
@@ -25,9 +24,6 @@ RUN echo "" > /env; \
        echo "export NO_ROCM_INST=1" >> /env; \
     fi
 
-#ENV TUNA_ROCM_VERSION=${OSDB_BKC_VERSION:+osdb-$OSDB_BKC_VERSION}
-#ENV TUNA_ROCM_VERSION=${TUNA_ROCM_VERSION:-rocm-$ROCMVERSION}
-
 RUN set -xe
 # Install dependencies
 RUN . /env; if [ -z $NO_ROCM_INST ]; then\
@@ -42,15 +38,6 @@ RUN . /env; if [ -z $NO_ROCM_INST ]; then\
             apt-get clean && \
             rm -rf /var/lib/apt/lists/*; \
     fi
-
-#FROM $BASEIMAGE as dtuna-ver-1
-#do nothing, assume rocm is installed here
-
-
-#FROM dtuna-ver-${ROCM_PRE} as final
-
-#finish building off previous image
-RUN set -xe
 
 # Install dependencies
 RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -f -y --allow-unauthenticated \
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 #default image to ubuntu + install rocm
-ARG BASEIMAGE=ubuntu:20.04
+ARG BASEIMAGE=rocm/miopen:ci_5450cc
 ARG ROCM_PRE=0
 #ARG IMG_VER=$([[ $BASEIMAGE == "ubuntu:20.04" ]]; echo $?)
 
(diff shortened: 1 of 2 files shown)


msaudulhassan:
Maybe in another PR we can use the [latest MIOpen commit](https://stackoverflow.com/q/45726013) and use a python/bash version of [`getDockerImageName()`](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/develop/Jenkinsfile#L172) to get the latest MIOpen docker image. I can give it a shot later

Dockerfile
```diff
@@ -1,49 +1,43 @@
...
+ARG BASEIMAGE=rocm/miopen:ci_5450cc
 
-FROM ubuntu:20.04 as dtuna-ver-0
+#FROM ubuntu:20.04 as dtuna-ver-0
+FROM $BASEIMAGE as dtuna-ver-0
 #install rocm
 ARG ROCMVERSION=
-ARG OSDB_BKC_VERSION='12969'
+ARG OSDB_BKC_VERSION=
+#'12969'
```
technicalgrp89, line 9: Do we need comment? line

Dockerfile
```diff
@@ -97,10 +91,10 @@ RUN wget https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.
 RUN dpkg -i dumb-init_*.deb && rm dumb-init_*.deb
 
 # Install cget
-RUN pip install cget
+#RUN pip install cget
```
technicalgrp89, line 94: Do we need this command under #?

Dockerfile
```diff
@@ -1,49 +1,43 @@
 #default image to ubuntu + install rocm
-ARG BASEIMAGE=ubuntu:20.04
-ARG ROCM_PRE=0
-#ARG IMG_VER=$([[ $BASEIMAGE == "ubuntu:20.04" ]]; echo $?)
+ARG BASEIMAGE=rocm/miopen:ci_5450cc
 
-FROM ubuntu:20.04 as dtuna-ver-0
+#FROM ubuntu:20.04 as dtuna-ver-0
```
technicalgrp89, line 4: Same as below observation

Dockerfile
cderb: It's just a note. BKC can take some time to look up. Though it has since been updated.

Dockerfile
cderb: The commands necessary for dependency install before the image was rebased to the miopen ci image have been left in, but disabled by comment.

Dockerfile
cderb: Left as a note of the typical ubuntu base image.

alexandraBara (approved):

technicalgrp89 (approved):

Dockerfile
```diff
@@ -1,49 +1,43 @@
...
+ARG OSDB_BKC_VERSION=
+#'12969'
+ENV NO_ROCM_INST=
 # Add rocm repository
-RUN apt-get update
-RUN apt-get install -y wget gnupg
+RUN apt-get update && apt-get install -y wget gnupg
 RUN wget -qO - http://repo.radeon.com/rocm/rocm.gpg.key | apt-key add -
-RUN if ! [ -z $OSDB_BKC_VERSION ]; then \
+RUN echo "" > /env; \
```
msaudulhassan, line 14: I'm still new to bash scripting. I wonder why `echo "" > /env;` is needed?

Dockerfile
```diff
@@ -1,49 +1,43 @@
...
        sh -c "echo deb [arch=amd64 trusted=yes] http://compute-artifactory.amd.com/artifactory/list/rocm-osdb-20.04-deb/ compute-rocm-dkms-no-npi-hipclang ${OSDB_BKC_VERSION} > /etc/apt/sources.list.d/rocm.list" ;\
        cat  /etc/apt/sources.list.d/rocm.list;\
-    else \
+    elif ! [ -z $ROCMVERSION ]; then \
        echo "Using Release VERSION: $ROCMVERSION";\
        sh -c "echo deb [arch=amd64 trusted=yes] http://compute-artifactory.amd.com/artifactory/list/rocm-osdb-20.04-deb/ compute-rocm-rel-${ROCMVERSION} > /etc/apt/sources.list.d/rocm.list" ;\
        cat  /etc/apt/sources.list.d/rocm.list;\
+    else \
+       echo "export NO_ROCM_INST=1" >> /env; \
     fi
```
msaudulhassan, line 25: Does this Dockerfile work if I provide OSDB_BKC_VERSION, like we used to do previously? I could be wrong, but it appears that here, you are making sure the Dockerfile maintains its previous functionality, but down below, the dependency installation has been changed, so the docker won't build.

Dockerfile
cderb: Each of the RUN statements is executed in a separate environment. To carry over env vars I'm passing around this file that can be executed at the start of a statement. However the file should exist or there will be a failure.

Dockerfile
cderb: The miopen dependencies are installed native to the miopen ci image at the top of the file, so we don't need to install them ourselves.
The rocm installation here is optional. It will work as before, but it is now possible to not install rocm. This allows for us to use the native rocm on the base image.

Dockerfile
msaudulhassan: Quick question: can we do `export ENV_VAR=123` for it to be carried over to all the child processes? We can do it in a separate PR if that's something that can help -- maybe prevent the build from crashing if the env file doesn't exist?

PR #940, opened 2023-11-11
Remove arch_full column from Machine and gather the same data in SessionRocMLIR where it's used

pcf000:
Since machine.gpus retains all the data that was parsed to set arch_full, it turned out to be easy to remove the arch_full field and do the same work in add_new_session.

### DIFF
--- a/tuna/machine.py
+++ b/tuna/machine.py
@@ -70,6 +70,7 @@ class Machine(BASE):  #pylint: disable=too-many-instance-attributes
   password: str = Column(Text, nullable=False)
   avail_gpus: List[int] = Column(Text, nullable=False)
   arch: str = Column(Text, nullable=False)
+  arch_full: str = Column(Text, nullable=False, default=arch)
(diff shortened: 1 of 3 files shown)


tuna/machine.py
```diff
@@ -70,7 +70,6 @@ class Machine(BASE):  #pylint: disable=too-many-instance-attributes
   password: str = Column(Text, nullable=False)
   avail_gpus: List[int] = Column(Text, nullable=False)
   arch: str = Column(Text, nullable=False)
-  arch_full: str = Column(Text, nullable=False, default=arch)
```
cderb, line 73: We can limit this to a one line change by changing this line to
```py
arch_full: str = ""
```

cderb (approved):

cderb (changes requested):
arch_full should be removed from addMachine in var/utils.groovy.

cderb (approved):

alexandraBara (approved):

PR #941, opened 2023-11-15
Enable conditional steps retrievals for evaluators 

reidkwja:
new -> compiled -> bar_lock

### DIFF
--- a/tuna/miopen/worker/fin_eval.py
+++ b/tuna/miopen/worker/fin_eval.py
@@ -281,9 +281,9 @@ class FinEvaluator(FinClass):
 
     if not self.get_job("compiled", "eval_start", True):
       if not self.get_job("new", "eval_start", True):
-          with self.bar_lock:
-              self.num_procs.value -= 1
-          return False
+        with self.bar_lock:
+          self.num_procs.value -= 1
+        return False
 
     orig_state = 'compiled'
     self.logger.info('Acquired new job: job_id=%s', self.job.id)
--- a/tuna/miopen/worker/fin_eval.py
+++ b/tuna/miopen/worker/fin_eval.py
@@ -54,10 +54,15 @@ class FinEvaluator(FinClass):
   def get_job(self, find_state, set_state, imply_end):
     """Polling to see if job available"""
     self.logger.info('find job: %s', find_state)
+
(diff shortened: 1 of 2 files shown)


tuna/miopen/worker/fin_eval.py
```diff
@@ -54,10 +54,12 @@ def __init__(self, **kwargs):
   def get_job(self, find_state, set_state, imply_end):
     """Polling to see if job available"""
     self.logger.info('find job: %s', find_state)
+
     if not super().get_job(find_state, set_state, imply_end):
       with self.bar_lock:
         self.num_procs.value -= 1
       return False
```
cderb: This should be evaluated after we've finished checking for all job times. After checking "compiled" and "new". Only when there are no more jobs.

cderb (changes requested):

alexandraBara:
What is the reason for implementing this? the job queue is going away completely once celery is fully implemented but I need to know what the use case is here, in case this is something i need to consider. Keep in mind that the entire job queue is being moved to celery -  which handles the implementation details for us.

cderb:
@alexandraBara We need the evaluation process to be able to retrieve jobs from the new state as well. We are also trying to prioritize compiled jobs.

tuna/miopen/worker/fin_eval.py
```diff
@@ -268,6 +259,25 @@ def close_job(self):
     self.set_job_state('evaluated')
     self.clean_cache_table()
 
+  def get_job(self, find_state, set_state, imply_end):
+    """Polling to see if job available"""
+    self.logger.info('find job: %s', find_state)
+
+    if not super().get_job(find_state, set_state, imply_end):
+      return False
+    return True
```
cderb, line 268: Safe to remove this function.

cderb (approved):

alexandraBara (approved):

tuna/miopen/worker/fin_eval.py
msaudulhassan, line 268: This can be replaced with
```python
return super().get_job(find_state, set_state, imply_end)
```

msaudulhassan (approved):
LGTM. Here's a minor suggestion though

PR #942, opened 2023-11-16
miopen specific modules de referencing from tuna folder level

technicalgrp89:
get miopen version method is moved in to miopen/worker folder.


tuna/worker_interface.py
```diff
@@ -398,7 +387,7 @@ def check_env(self) -> bool:
       raise ValueError(
           f'session rocm_v {self.dbt.session.rocm_v} does not match env rocm_v {env_rocm_v}'
       )
-    env_miopen_v: str = self.get_miopen_v()
+    env_miopen_v: str = chk_miopen_env_v(self)
```
cderb: miopen_v should not be present in worker_interface.

tuna/miopen/worker/fin_class.py
```diff
@@ -126,20 +126,6 @@ def __init__(self, **kwargs):
...
-        "cat /opt/rocm/include/miopen/version.h "
-        "| grep MIOPEN_VERSION_TWEAK | cut -d ' ' -f 3")
-    if "No such file" in commit_hash:
-      _, commit_hash, _ = self.exec_docker_cmd(
-          "cat /opt/rocm/miopen/include/miopen/version.h "
-          "| grep MIOPEN_VERSION_TWEAK | cut -d ' ' -f 3")
-    self.logger.info('Got branch commit hash: %s', commit_hash)
-
-    return commit_hash
-
```
cderb: This function should stay in fin_class.

tuna/miopen/worker/fin_utils.py
```diff
@@ -156,3 +157,24 @@ def get_tensor(tensor_type: str, tensor_dict: dict) -> dict:
...
+        "| grep MIOPEN_VERSION_TWEAK | cut -d ' ' -f 3")
+  self.logger.info('Got branch commit hash: %s', commit_hash)
+  return commit_hash
+
+
+def chk_miopen_env_v(self) -> str:
+  """Interface function to get new branch hash"""
+  env_miopen_v: str = self.get_miopen_v()
+
+  return env_miopen_v
```
cderb: No need for these functions.

tuna/miopen/db/session.py
```diff
@@ -71,6 +72,6 @@ def add_new_session(self, args, worker):
     if hasattr(args, 'miopen_v') and args.miopen_v:
       self.miopen_v = args.miopen_v
     else:
-      self.miopen_v = worker.get_miopen_v()
+      self.miopen_v = get_miopen_v(self)
```
cderb: The worker that is submitted to this function needs to be a derivative of FinClass. Specify that in the type hint, and leave the call to worker.get_miopen_v.

cderb (changes requested):

tuna/worker_interface.py
technicalgrp89: I will rename the API name. which would be referred from fin_class. pls confirm

cderb: You can remove this line, if there is some feature elsewhere in this class that refers to env_miopen_v, then that will need to be addressed as well.

tuna/worker_interface.py
```diff
@@ -65,6 +65,7 @@ class WorkerInterface(Process):
   # pylint: disable=too-many-instance-attributes
   # pylint: disable=too-many-public-methods
   # pylint: disable=too-many-statements
+  # pylint: disable=no-member
```
alexandraBara: can this be removed? which object is no-member?

technicalgrp89:
Created separate branch with all changes.

PR #944, opened 2023-11-20
Catching up on rocMLIR changes:  quick-tuning, attention op, backward data, small robustness improvements

pcf000:
Tuning-space support, enable backward-data tuning.
Tuning space as an Enum.
Improving exception reporting.
Use stdout instead of a results file for tuningRunner.py.
Support for attention op.
Bail without updating database if tuning data is clearly bogus, to avoid database barfs.
Initialise arch_full to empty string;  __init__ will default it to arch if needed.

### DIFF
--- a/tuna/rocmlir/rocmlir_lib.py
+++ b/tuna/rocmlir/rocmlir_lib.py
@@ -81,6 +81,7 @@ class RocMLIR(MITunaInterface):
         help='Which space of tuning configs should be used while tuning')
 
     group: argparse._MutuallyExclusiveGroup = parser.add_mutually_exclusive_group(
+      required=True
     )
     group.add_argument('--add_tables',
                        dest='add_tables',
@@ -89,7 +90,6 @@ class RocMLIR(MITunaInterface):
 
     # pylint: disable=duplicate-code
     group.add_argument(
-        '-e',
         '--execute',
         dest='execute',
         action='store_true',
@@ -185,8 +185,8 @@ class RocMLIR(MITunaInterface):
         SessionRocMLIR().add_new_session(self.args, worker)
       return None
 
-    assert self.args.execute, \
-      "one of --add_tables, --init_session, or --execute must be present"
+    # Must be --execute, because the mutually-exclusive-group argument is
+    # required, and we just checked for --add_tables and --init_session.
     res = self.compose_worker_list(machines)
     return res
 
--- a/tuna/rocmlir/rocmlir_lib.py
+++ b/tuna/rocmlir/rocmlir_lib.py
@@ -73,6 +73,12 @@ class RocMLIR(MITunaInterface):
                         help='How many workers per GPU',
                         default=1.0,
                         type=float)
+    parser.add_argument(
+        '--tuning_space',
+        dest='tuning_space',
+        default='exhaustive',
+        choices=['quick', 'full', 'exhaustive'],
+        help='Which space of tuning configs should be used while tuning')
 
     group: argparse._MutuallyExclusiveGroup = parser.add_mutually_exclusive_group(
     )
@@ -153,8 +159,10 @@ class RocMLIR(MITunaInterface):
     # pylint: disable=duplicate-code
     """Generates the library specific schema to the connected SQL server."""
     ret_t: bool = create_tables(get_tables())
-    recreate_triggers(['conv_timestamp_trigger', 'gemm_timestamp_trigger'],
-                      get_timestamp_trigger())
+    recreate_triggers([
+        'conv_timestamp_trigger', 'gemm_timestamp_trigger',
+        'attention_timestamp_trigger'
+    ], get_timestamp_trigger())
     self.logger.info('DB creation successful: %s', ret_t)
     return True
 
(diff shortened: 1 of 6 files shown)


tuna/rocmlir/rocmlir_worker.py
```diff
@@ -170,17 +173,21 @@ def run_cmd(self):
...
 
     if not os.path.exists("./bin/tuningRunner.py"):
       raise FileNotFoundError("tuningRunner.py not found;"
                               "  wrong directory or missing setup")
 
     cmd = env_str + f" python3 ./bin/tuningRunner.py -q {special_args} \
                      --config='{config_string}' --mlir-build-dir `pwd` \
-                     --output={self.output_filename()} --tflops \
-                     --rocmlir_gen_flags='--device={self.gpu_id}'"
+                     --output=- --tflops \
```
pcf000, line 189: '-' represents stdout, which tuningRunner.py now supports for its output.  Safer than writing to a file because we don't need to add hoops to avoid collisions.  Redirect stderr out of the way, because otherwise it confuses things.

tuna/rocmlir/rocmlir_lib.py
```diff
@@ -177,6 +185,8 @@ def run(self) -> Optional[List[RocMLIRWorker]]:
         SessionRocMLIR().add_new_session(self.args, worker)
       return None
 
+    assert self.args.execute, \
```
alexandraBara: args should probably be checked at the end of the parse_args function, otherwise you end up with random arg checks in code and it will become hard to track down.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -579,6 +585,183 @@ class GEMMResults(BASE, ResultsMixin):  # pylint: disable=too-many-instance-attr
...
+    """Return config as a flag/value string suitable for tuningRunner.py."""
+    string = ""
+    #     for field, value in self.to_dict().items():
+    #       flag = self.options[field]
+    #       if flag:
+    #         string += f"{flag} {value} "
+    # In options order for canonicalisation, kind of.
+    for field, flag in self.options.items():
+      value = getattr(self, field, None)
+      if value is not None and flag is not None:
```
alexandraBara, line 640:
```suggestion
      if value and flag:
```

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -579,6 +585,183 @@ class GEMMResults(BASE, ResultsMixin):  # pylint: disable=too-many-instance-attr
...
+        '-with-attn-scale': 'with_attn_scale',
+        '-t': 'data_type'
+    }
+    # kernel-repeats has no flag, but perfRunner.py uses 5.
+
+    self.kernel_repeats = 1
+    for flag, value in options.items():
+      if value == "true":
+        value = 1
+      if value == "false":
```
alexandraBara: can it be something other than true/false? if so there needs to be a 3rd default case, otherwise this needs to be modified to if value==true-> set to 1, else set to 0. Once you start introducing type hints, it will complain.

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -579,6 +585,183 @@ class GEMMResults(BASE, ResultsMixin):  # pylint: disable=too-many-instance-attr
...
+        value = 0
+      field = fields[flag]
+      if field:
+        setattr(self, field, value)
+
+  ## Adapted from perfRunner.getGemmConfigurations.
+
+  DATA_TYPES = ['f32', 'f16']
```
alexandraBara: ideally macros go at the top of the page

tuna/rocmlir/rocmlir_tables.py
```diff
@@ -579,6 +585,183 @@ class GEMMResults(BASE, ResultsMixin):  # pylint: disable=too-many-instance-attr
...
+
+  def get_configurations(self, filename):
+    #pylint: disable=invalid-name
+    #pylint: disable=too-many-locals
```
alexandraBara: this is definitely not one to disable, this function is too large, it needs to be broken down in multiple smaller functions, code is otherwise hard to track.

cderb (approved):

alexandraBara (changes requested):
Looks like i forgot to hit submit on this review

tuna/rocmlir/rocmlir_tables.py
pcf000: Value might exist but be 0 or False, and I specifically want to check for non-existence.  Flag will be either a string or absent, but I liked the parallel construct.

tuna/rocmlir/rocmlir_lib.py
pcf000: If I make the _MutuallyExclusiveGroup that includes these three required, then parse_args itself enforces that one of them is set and I don't need the assert at all.  Update coming.

tuna/rocmlir/rocmlir_tables.py
pcf000: Value can also be a number, or a string like 'i8' or 'f32'.  The tests here are to canonicalise bools to 1 and 0.  I'm nearly convinced that they'll always be lowercase, but perfRunner.py accepts either case and I'm changing this spot to do the same.

alexandraBara (approved):
LGTM and this file is linted correct?

alexandraBara:
@msaudulhassan there is a CI fail here with the SolverAnalytics repo, can you please take a look and help with fixing

PR #946, opened 2023-11-30
Batched Load Jobs

msaudulhassan:
load_jobs first tries to do a bulk insert into the database with all the jobs. This can fail in two cases: (1) there are too too many jobs, (2) some of the jobs to be inserted already exist in the database. If bulk insert fails, load_job reverts to inserting one job at a time, which is slow.

Proposed change:

If bulk commit fails, revert to "batched commits," wherein batches of jobs (say 1000 jobs at a time) are insert into the database. By using the `IGNORE` keyword, duplicates can be handled.

### DIFF
--- a/tuna/miopen/subcmd/load_job.py
+++ b/tuna/miopen/subcmd/load_job.py
@@ -27,6 +27,7 @@
 """
 Script for adding jobs to the MySQL database
 """
+import pdb
 import logging
 import argparse
 import warnings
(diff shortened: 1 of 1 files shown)


tuna/miopen/subcmd/load_job.py
```diff
@@ -27,6 +27,7 @@
 """
 Script for adding jobs to the MySQL database
 """
+import pdb
```
cderb, line 30: cleanup

tuna/miopen/subcmd/load_job.py
```diff
@@ -163,43 +164,62 @@ def add_jobs(args: argparse.Namespace, dbt: MIOpenDBTables,
...
+
+    try:
+      session.commit()
+    except IntegrityError as err:
+      session.rollback()
+      counts = 0
+      logger.warning(
+          'Quick update failed, trying batched update : %s', err)
+
+      print('\n')
```
cderb, line 192: cleanup

cderb (approved):
Remove debug lines, otherwise LGTM.

alexandraBara (changes requested):
Please add this file to pylint, if it were linted it would complain that pdb import is not used, code looks good

alexandraBara:
Stale

PR #947, opened 2023-12-11
Circular dependency , MIOpen specific code re ordering

technicalgrp89:
tuna/miopen/db
tuna/miopen/db/worker
folders are resolved for cyclic imports. 
-  New table/class specific files are created


tuna/worker_interface.py
```diff
@@ -134,6 +134,9 @@ def __init__(self, **kwargs):
     #also set cnx here in case WorkerInterface exec_command etc called directly
     self.cnx: Connection = self.machine.connect(chk_abort_file)
 
+  def get_miopen_v(self):
+    """Abstract method """
```
alexandraBara: should raise NotImplemented otherwise you are only writing this to pass pylint.

tests/test_importconfigs.py
```diff
@@ -40,7 +40,7 @@
 from tuna.miopen.db.tables import MIOpenDBTables, ConfigType
 from utils import CfgImportArgs
 from tuna.miopen.db.benchmark import Framework, ModelEnum, FrameworkEnum
-from tuna.miopen.db.miopen_tables import ConvolutionBenchmark
+from tuna.miopen.db.convolutionjob_tables import ConvolutionBenchmark
```
alexandraBara, line 43: I do like the idea of splitting these tables out into their own files

alexandraBara (changes requested):

tests/test_importconfigs.py
technicalgrp89: All the tables which are used in tables.py are creating circular dependency. We need to re arrange them to resolve the same.

tuna/miopen/db/__init__.py
```diff
@@ -0,0 +1,25 @@
+###############################################################################
```
cderb: Why add a license to a blank file?

tuna/worker_interface.py
cderb: This method needs to be removed from worker_interface.

tests/test_add_session.py
```diff
@@ -79,7 +80,7 @@ def test_add_session():
   args.docker_name = docker_name
   args.solver_id = 1
 
-  worker = WorkerInterface(**kwargs)
+  worker = FinClass(**kwargs)
```
cderb, line 83: This should still be possible with WorkerInterface, why use a child class?

cderb (changes requested):

tests/test_add_session.py
technicalgrp89: sess_id = Session().add_new_session(args, worker), in implementation we are passing as worker:finClass. this leads to empty WorkerInterace Object creation. As worker is not FinClass type. We have to initialize is to worker = FinClass(**kwargs)
in Session.py
def add_new_session(self, args, worker: FinClass): , to support following. 
 if hasattr(args, 'miopen_v') and args.miopen_v:
      self.miopen_v = args.miopen_v
    else:
      self.miopen_v = worker.get_miopen_v()
worker class does not support get_miopen_v() it is moved to finclass

tuna/miopen/db/__init__.py
technicalgrp89: To avoid the import fail error - in submodules.

cderb: Leave the file, remove the license.

tests/test_add_session.py
cderb: I see, this appears to be an miopen test.

cderb (approved):
lgtm

alexandraBara (approved):

PR #950, opened 2024-01-12
New models

alexandraBara:
Added new models during tuning, they are enums so they need definition updates here. The production DB is already updated.


cderb (approved):

technicalgrp89 (approved):
lgtm

PR #954, opened 2024-02-16
Celery intergration

alexandraBara:
Switch find_compile/eval and perf_compile/eval to run through celery


tuna/example/example_worker.py
```diff
@@ -51,7 +51,7 @@ def set_db_tables(self) -> None:
   def step(self) -> bool:
     """Main functionality of the worker class. It picks up jobs in new state and executes them"""
 
-    if not self.get_job("new", "running", False):
+    if not self.get_job({"new"}, "running", False):
```
cderb, line 54: Curly braces would indicate a dictionary in python. This function hints that the find_state parameter is a list.

tuna/miopen/miopen_lib.py
```diff
@@ -395,19 +400,213 @@ def get_envmt(self):
...
+      job_attr = [column.name for column in inspect(self.dbt.job_table).c]
+      job_attr.remove("insert_ts")
+      job_attr.remove("update_ts")
+    except NoInspectionAvailable as error:
+      self.logger.warning("Ignoring error for init_session: %s", error)
+    return job_attr
+
+  def get_jobs(self,
+               session: DbSession,
+               find_state: List[str],
```
cderb, line 430: Good this is expanded to list, saves us 1 query in eval.

tuna/worker_interface.py
```diff
@@ -91,9 +93,12 @@ def __init__(self, **kwargs):
     self.end_jobs = None
     #job detail vars
     self.envmt: List = []
-    self.fetch_state: List = ['new']
+    self.fetch_state = set()
```
cderb, line 96: Set is a fair idea for this var. The type hint for get_job should be updated to reflect this.

cderb:
LGTM, would like to run a scale test before approving.

tuna/example/example_worker.py
alexandraBara: ive made it a dict because it later translates to mysql nicely for querring for jobs where the state is in (). i can look into changing to a list or something like that

PR #958, opened 2024-02-23
Applicability generation performance update

cderb:
Reduce processes generated for low applicability workloads, improve logging

### DIFF
--- a/tuna/miopen/worker/fin_class.py
+++ b/tuna/miopen/worker/fin_class.py
@@ -381,9 +381,8 @@ class FinClass(WorkerInterface):
 
       block_size = len_rows // num_blk  #size of the config block
       extra = len_rows % num_blk  #leftover configs, don't divide evenly
-      self.logger.info(
-          "cfg workdiv: num_blocks: %s, block_size: %s, extra: %s", num_blk,
-          block_size, extra)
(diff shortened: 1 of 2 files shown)


tuna/miopen/worker/fin_class.py
```diff
@@ -348,52 +348,57 @@ def __rm_old_app(self, session: DbSession, cfg_rows: list) -> None:
...
-          r_dict = compose_config_obj(row, self.config_type)
-          if self.config_type == ConfigType.batch_norm:
-            r_dict['direction'] = row.get_direction()
-          master_cfg_list.append(r_dict)
+      len_rows = len(rows)
+      master_cfg_list = []
+      for row in rows:
+        r_dict = compose_config_obj(row, self.config_type)
+        if self.config_type == ConfigType.batch_norm:
+          r_dict['direction'] = row.get_direction()
```
alexandraBara, line 376: can this be moved inside compose_config_obj(), otherwise it will have to be accounted for separately which will definitely lead to oversights

alexandraBara (changes requested):

tuna/miopen/worker/fin_class.py
cderb: out of scope

alexandraBara (approved):

reidkwja (approved):

PR #960, opened 2024-02-23
Minor bug fixes

cderb:
opencl binary location
duplicate miopen table creation
missing layouts from config driver
job step query

### DIFF
--- a/tuna/miopen/driver/base.py
+++ b/tuna/miopen/driver/base.py
@@ -214,7 +214,7 @@ class MIOpenDriver(DriverBase):
       i_dict['dim3'] = self.in_h
       i_dict['dim4'] = self.in_w
       i_dict['layout'] = self.in_layout
-    elif self.in_layout == 'NHWC':
+    elif self.in_layout in ('NHWC', 'NDHWC'):
       i_dict['dim1'] = self.in_d
       i_dict['dim2'] = self.in_h
       i_dict['dim3'] = self.in_w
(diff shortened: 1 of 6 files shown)


tuna/miopen/driver/base.py
```diff
@@ -214,7 +214,7 @@ def __compose_input_t(self) -> Dict[str, int]:
       i_dict['dim3'] = self.in_h
       i_dict['dim4'] = self.in_w
       i_dict['layout'] = self.in_layout
-    elif self.in_layout == 'NHWC':
+    elif self.in_layout in ('NHWC', 'NDHWC'):
```
alexandraBara, line 217: i see us failing to catch this everytime a new layout is added. Maybe we can add a map with supported layouts in the metadata and have that be the only place to edit when a new layout is added. We can have an supported in_layout, fil and weight? It should also be checked when we parse a driver cmd to insert into the db. That way we catch unsupported layouts before we add to the DB, as well as when we read from DB.

cderb: Yes, this could be better handled in a function referring to a central definition.

alexandraBara: Should we create a ticket or do you want to update the code in this PR?

cderb: I added a check for supported layouts. Further work to consolidate layout handling would prove to be non-trivial and would be better handled in a PR with that focus.

alexandraBara (approved):

reidkwja (approved):

PR #962, opened 2024-03-27
Celery intro for job scheduling

alexandraBara:
Using celery as the scheduler for the following tuning runs:

- miopen_perf_compile/eval
- fin_find_compile/eval

Celery workers are automatically launched when tuning runs.
New dependency added: redis

celery folder: /tuna/celery_app
miopen celery folder: /tuna/miopen/celery_tuning


tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,377 @@
...
+
+    if failed_job:
+      set_job_state(session, job, dbt, 'errored', False, result=result_str)
+    else:
+      set_job_state(session, job, dbt, 'compiled', False, result=result_str)
+
+  return True
+
+
+def process_fin_evaluator_results(fin_json, context, dbt):
```
alexandraBara: @cderb  need eyes on these process functions

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,377 @@
...
+def tune(library, job_batch_size=1000):
+  """tuning loop to spin out celery tasks"""
+
+  tune_start_t = time.time()
+  worker_type, kwargs, fdb_attr, q_name = prep_tuning(library)
+  res_set = ResultSet([])
+  job_start_t = None
+  job_end_t = None
+
+  while True:
```
alexandraBara: All jobs get pulled from mySQL and enqueued to redis

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,377 @@
...
+                'config': config,
+                'worker_type': worker_type,
+                'arch': library.dbt.session.arch,
+                'num_cu': library.dbt.session.num_cu,
+                'kwargs': kwargs,
+                'fdb_attr': fdb_attr
+            }
+
+            #calling celery task, enqueuing to celery queue
+            res_set.add(hardware_pick.apply_async((context,), queue=q_name))
```
alexandraBara: celery async task launch

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,377 @@
...
+  drain_process = Process(target=drain_queue, args=[worker_type, DBT])
+  drain_process.start()
+
+  LOGGER.info('Gathering async results in callback function, blocking')
+  #waiting on all results to come in
+  _ = res_set.join(callback=result_callback)
+
+  #terminate result_queue drain process with special queue token (NONE,NONE)
+  LOGGER.info('Adding terminatig token to result_queue')
+  result_queue.put([None, None])
```
alexandraBara: launch separate result queue drain that writes back to mySQL, blocking res_set join that waits for all enqueued jobs to terminate

tuna/miopen/utils/json_to_sql.py
```diff
@@ -0,0 +1,358 @@
...
+from tuna.utils.db_utility import session_retry, gen_select_objs
+from tuna.utils.db_utility import gen_update_query, gen_insert_query
+from tuna.miopen.worker.fin_utils import get_fin_slv_status
+from tuna.miopen.utils.parsing import parse_pdb_key
+from tuna.miopen.db.solver import get_solver_ids
+
+LOGGER = setup_logger('parse_results')
+
+
+def __update_fdb_w_kernels(  #pylint: disable=too-many-arguments,too-many-locals
```
alexandraBara, line 43: file that contains all the fin results processing functions moved over from fin_builder/eval

tuna/celery_app/celery_app.py
```diff
@@ -0,0 +1,75 @@
...
+#
+###############################################################################
+"""Module to define celery app"""
+import os
+import subprocess
+from celery import Celery
+from celery.utils.log import get_task_logger
+
+TUNA_CELERY_BROKER = 'mituna_redis'
+TUNA_REDIS_PORT = '6379'
```
msaudulhassan: Could you brief these variables in the README

tuna/celery_app/celery_app.py
```diff
@@ -0,0 +1,75 @@
...
+
+app.conf.update(result_expires=3600,)
+app.autodiscover_tasks()
+app.conf.result_backend_transport_options = {'retry_policy': {'timeout': 5.0}}
+
+
+def stop_active_workers():
+  """Shutdown active workers"""
+  if app.control.inspect().active() is not None:
+    app.control.shutdown()
```
msaudulhassan, line 63: maybe also call `app.control.revoke`? See https://stackoverflow.com/a/49521560

tuna/go_fish.py
```diff
@@ -69,39 +71,46 @@ def parse_args() -> Dict[str, Any]:
...
   yaml_files = [args['yaml']]
 
   #case with yaml file
   if args['yaml']:
     yaml_files = parse_yaml(args['yaml'], args['lib'])
 
-  worker_lst: list
+  job_batch_size = 1000
+  if 'TUNA_CELERY_JOB_BATCH_SIZE' in os.environ:
+    job_batch_size = int(os.environ['TUNA_CELERY_JOB_BATCH_SIZE'])
```
msaudulhassan, line 92: a short note on `TUNA_CELERY_JOB_BATCH_SIZE` in README would be helpful

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,395 @@
...
+from tuna.miopen.worker.fin_utils import get_fin_result
+from tuna.miopen.db.solver import get_solver_ids
+from tuna.celery_app.celery_workers import launch_celery_worker
+<<<<<<< HEAD
+from tuna.miopen.celery_tuning.celery_tasks import celery_enqueue
+from tuna.celery_app.celery_app import stop_active_workers
+=======
+from tuna.miopen.celery_tuning.celery_tasks import hardware_pick
+from tuna.celery_app.celery_app import stop_active_workers, purge_queue
+>>>>>>> alex_celery4
```
msaudulhassan: merging mess :)

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,395 @@
...
+      #stop_active_workers()
+      if not launch_celery_worker(library, q_name, machines,
+                                  get_worker_granularity(library)):
+        raise ValueError('Could not launch celery worker')
+    except kombu.exceptions.OperationalError as k_err:
+      LOGGER.error('Redis error ocurred: %s', k_err)
+      return False
+
+  global DBT  #pylint: disable=global-variable-undefined
+  DBT = MIOpenDBTables(session_id=library.args.session_id,
```
msaudulhassan, line 255: can we do away with this global var?

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,390 @@
...
+                                    job_batch_size)
+
+        for i in range(0, len(job_list), job_batch_size):
+          batch_jobs = job_list[i:min(i + job_batch_size, len(job_list))]
+          if library.args.fin_steps:
+            entries = library.compose_work_objs_fin(session, batch_jobs,
+                                                    library.dbt)
+
+          entries = library.compose_work_objs_fin(session, job_list,
+                                                  library.dbt)
```
cderb: Appears to always overwrite entries. Desired behavior?

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,390 @@
...
+          LOGGER.info('All tasks added to queue')
+          break
+      except KeyboardInterrupt:
+        LOGGER.error('Keyboard interrupt caught, draining results queue')
+        session.rollback()
+        stop_active_workers()
+        results_gather(res_set, worker_type)
+        purge_queue([q_name], LOGGER)
+
+  results_gather(res_set, worker_type)
```
cderb: Will be preferable to have the async result gathering started before Tuna begins to fill the celery queue. So that Tuna isn't waiting on all work to be added before it starts processing results.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,390 @@
...
+  drain_process.join()
+  LOGGER.info('Done writing from result_queue to mySQL')
+
+
+def drain_queue(worker_type, dbt):
+  """Drain results queue"""
+  interrupt = False
+  LOGGER.info('Draining queue')
+  sleep_time = 0
+  wait_limit = 1800  #max wait time
```
cderb: This (and lines 379-388) will terminate drain_queue after 30 minutes of accumulated deep sleep time?
Deep sleep being periods of over 2s where no work is completed (with an incremental use of 2s).
This seems generous, but I can also foresee the case where this could be running for a week and the accumulated delays would cause drain_queue to terminate before emptying the queue.
The bigger the workload, the higher the probability that all processors with find jobs that keep them busy for more than 2s.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,390 @@
...
+          process_fin_builder_results(session, fin_json, context, dbt)
+          seen_res = True
+        else:
+          process_fin_evaluator_results(session, fin_json, context, dbt)
+          seen_res = True
+      except KeyboardInterrupt:
+        if result_queue.empty():
+          return True
+        interrupt = True
+        #continue
```
cderb: In the case where Tuna is terminated early via interrupt. This would not stop until the result queue is empty and a second keyboard interrupt is issued. Or it could expire from exceeding the wait_limit.
It seems drain_queue relies on wait_limit for it to exit gracefully after Tuna is interrupted.
At the same time wait_limit threatens Tuna's result commits with early termination for long runs.
Would the result queue persist after this process ends? And would it be easy / reliable to restart the enqueueing process to resume committing the results.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,395 @@
...
+                                    job_batch_size)
+
+        for i in range(0, len(job_list), job_batch_size):
+          batch_jobs = job_list[i:min(i + job_batch_size, len(job_list))]
+          if library.args.fin_steps:
+            entries = library.compose_work_objs_fin(session, batch_jobs,
+                                                    library.dbt)
+
+          entries = library.compose_work_objs_fin(session, job_list,
+                                                  library.dbt)
```
msaudulhassan: If I'm not wrong, you seem to be processing the whole job_list again?

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,395 @@
...
+
+  LOGGER.info('Gathering async results in callback function, blocking')
+  #waiting on all results to come in
+  _ = res_set.join(callback=result_callback)
+
+  #terminate result_queue drain process with special queue token (NONE,NONE)
+  LOGGER.info('Adding terminatig token to result_queue')
+  result_queue.put([None, None])
+
+  drain_process.join()
```
msaudulhassan: Do you monitor the number of `[None, None]` tokens to determine if all threads have finished? I believe you mentioned that you just monitor if there are any unfinshed jobs left, but just to confirm again...

msaudulhassan (commented):
Partial review

tuna/miopen/celery_tuning/tuning.py
alexandraBara: this is in the main process so there is only 1 token of this kind being added to the Q. the minute we see it, we terminate waiting on results because all the results have already come in

tuna/miopen/celery_tuning/tuning.py
alexandraBara: correct, @cderb  caught this as well, ill address

tuna/miopen/celery_tuning/tuning.py
alexandraBara: thanks for catching, not intended behavior

tuna/miopen/celery_tuning/tuning.py
alexandraBara: correct, i have changes to push where this is moved before the get_jobs gets called.

tuna/miopen/celery_tuning/tuning.py
alexandraBara: yes that was the intended behavior, to have a max wait time. As we discussed, I will find a different way to wait. I can up the wait time to something like 1 hour if that is more reasonable.

tuna/miopen/celery_tuning/tuning.py
msaudulhassan: The `results_gather` function first launches drain_queue, and then puts [None, None] onto the queue. Is it possible that the queue might get drained even before a [None, None] is put there?

msaudulhassan: I'm not really familiar with multithreading, but does results_gather launch multiple drain_queue processes in parallel? Because in [tuning.py#L361](https://github.com/ROCm/MITuna/blob/2751d98346e1789823b4a013517d12ae707c8c7c/tuna/miopen/celery_tuning/tuning.py#L361), drain_queue uses the `get` function to remove an element off the queue and it then checks if the element is [None, None]. If it is, it returns. But since [None, None] has now been removed from the queue, the other drain_queue threads would will not hit the terminating condition.

tuna/miopen/celery_tuning/tuning.py
msaudulhassan: I see it's already fixed

tuna/miopen/celery_tuning/tuning.py
cderb: If we replace this with a completion patience, that may work better.
We could reduce the wait_limit to 10 minutes, but refresh it every time a job is received by the result queue.
Then the behavior would be if no job completes in 10 minutes, exit the process.
I think this would reduce the probability of early termination proportionately.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -0,0 +1,390 @@
...
+          #Note: reset job state for those in flight that have not reached the res_queue yet
+          return True
+        LOGGER.warning(exc)
+        LOGGER.info('Sleeping for 2 sec, waiting on results from celery')
+        if not seen_res:
+          wait_limit -= 2
+          if max(0, wait_limit) == 0:
+            LOGGER.info(
+                'Max wait limit of %s seconds has been reached, terminating...',
+                wait_limit)
```
cderb: decrement on wait_limit, logger always prints 0.

alexandraBara: im rewriting this wait code, stay tuned

tuna/miopen/celery_tuning/tuning.py
alexandraBara: there is only 1 drain_queue thread (this part is not multithreaded), the results Q that we drain is not shared by multiple processes. The none, none token is added to the results Q, after we do a thread join on the the threads that run the actual work and that works gets added to the results Q. Since join is blocking, [none, none] gets added at the end of the queue after all results have come in. When we dequeue and see the token that means we have reached the end of the Q.

tuna/miopen/celery_tuning/tuning.py
alexandraBara: i tried, i did not find a way to get rid of it. Information from here is needed before and after celery tasks have run and since this structure is not serializable, we cannot pass it to celery.

tuna/go_fish.py
alexandraBara: im going to have to rewrite a lot of README to add notes on celery. Ill make this a task to update after im done with all the celery updates (so I can add a README on the final version)

tuna/miopen/celery_tuning/tuning.py
msaudulhassan: I see.

tuna/mituna_interface.py
```diff
@@ -184,20 +203,48 @@ def get_kwargs(self, gpu_idx: int, f_vals: Dict[str, Any]) -> Dict[str, Any]:
...
+      kwargs["num_procs"] = f_vals["num_procs"]
+      kwargs["barred"] = f_vals["barred"]
+      kwargs["bar_lock"] = f_vals["bar_lock"]
+      kwargs["end_jobs"] = f_vals["end_jobs"]
+      kwargs["job_queue"] = f_vals["job_queue"]
+      kwargs["job_queue_lock"] = f_vals["job_queue_lock"]
+
     return kwargs
+
+  def get_jobs(self, session: DbSession, find_state: List[str], set_state: str,
```
reidkwja, line 225: There are many get_jobs, gets_kwargs across files in this PR...It might be a good idea to either aggregate them all into a class and inherit a variety of it [using some design paatern] or name each function with more specific name` get_<specialtask>_job():`

tuna/celery_app/celery_app.py
```diff
@@ -0,0 +1,79 @@
...
+###############################################################################
+"""Module to define celery app"""
+import os
+import subprocess
+from celery import Celery
+from celery.utils.log import get_task_logger
+
+#TUNA_CELERY_BROKER = 'mituna_redis'
+TUNA_CELERY_BROKER = 'localhost'
+TUNA_REDIS_PORT = '6379'
```
reidkwja, line 35: For both the broker and port, it might be a much flexible idea to pass it as environmental variable or parse it from  .ini config file

reidkwja (changes requested):

tuna/celery_app/celery_app.py
alexandraBara: these are just the defaults here, later i overwrite them with whats in the env

tuna/mituna_interface.py
alexandraBara: you are correct, ill take a look to see if there is a pattern we can put in place (beyond scope of this PR though)

Dmantri98 (approved):

tuna/celery_app/celery_app.py
```diff
@@ -40,14 +40,17 @@
...
   TUNA_REDIS_PORT = os.environ['TUNA_REDIS_PORT']
 
-app = Celery('celery_app',
-             broker_url=f"redis://{TUNA_CELERY_BROKER}:{TUNA_REDIS_PORT}//",
-             result_backend=f"redis://{TUNA_CELERY_BROKER}:{TUNA_REDIS_PORT}/",
-             include=['tuna.miopen.celery_tuning.celery_tasks'])
+app = Celery(
+    'celery_app',
+    broker_url=f"redis://{TUNA_CELERY_BROKER}:{TUNA_REDIS_PORT}/14",
+    result_backend=f"redis://{TUNA_CELERY_BROKER}:{TUNA_REDIS_PORT}/15",
```
JehandadKhan, line 46: Please dont hard code the db numbers, they should at least come from a config!

tuna/miopen/celery_tuning/celery_tasks.py
```diff
@@ -49,4 +49,4 @@ def celery_enqueue(context):
   logger.info("Enqueueing task %s", context['job'])
   worker = prep_worker(copy.deepcopy(context))
   ret = worker.run()
-  return ret, context
+  return {"ret": ret, "context": context}
```
JehandadKhan, line 52: Perhaps this was not necessary, since the error seemed to stem from the sharing of databases for the broker and the result backend

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -290,102 +291,75 @@ def tune(library, job_batch_size=1000):
       subp.kill()
     return False
 
-  res_set = ResultSet([])
+  #res_set = ResultSet([])
```
JehandadKhan, line 298: No longer needed.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -290,102 +291,75 @@ def tune(library, job_batch_size=1000):
...
-        purge_queue([q_name])
-        library.cancel_consumer(q_name)
-
-  results_gather_terminate(res_set, drain_process)
-  library.cancel_consumer(q_name)
-  end = time.time()
-  LOGGER.info("Took {:0>8} to tune".format(str(timedelta(seconds=end - start))))  #pylint: disable=consider-using-f-string
-  LOGGER.info("{:0>8} of which was spent enqueuing jobs".format(  #pylint: disable=consider-using-f-string
-      str(timedelta(seconds=j_end - start))))
+    job_list = library.get_jobs(session, library.fetch_state,
```
JehandadKhan: make sure there is logging in this function to indicate progress.

JehandadKhan: Perhaps the query can be simplified so that its a `SELECT count(*)` instead of `select *`.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -290,102 +291,75 @@ def tune(library, job_batch_size=1000):
...
-  LOGGER.info('fin_json : %s', value[0])
-  LOGGER.info('context : %s', value[1])
-  #result, v = app.AsyncResult(task_id).get()
-  #LOGGER.info('v: %s', v)
-  #LOGGER.info('results: %s', result)
-
-  result_queue.put([value[0], value[1]])
+  except KeyboardInterrupt:
+    LOGGER.error('Keyboard interrupt caught, draining results queue')
+    purge_queue([q_name])
```
JehandadKhan, line 336: If you dont purge the queue, you would not have to requeue when you restart. However, that might make state management more complicated.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -290,102 +291,75 @@ def tune(library, job_batch_size=1000):
...
-  LOGGER.info('Started drain process')
-  #start draining result_queue in subprocess
-  event = threading.Event()
-  drain_process = Process(target=drain_queue, args=[worker_type, DBT, event])
-  drain_process.start()
-
-  return drain_process
 
+  try:
+    enqueue_proc = Process(target=enqueue_jobs, args=[library, job_batch_size, worker_type, kwargs, fdb_attr, q_name])
```
JehandadKhan: When I was testing this, process seemed to be the easiest thing to do, later however, I felt we can stick to `async` for everything, however, I dint put much thought into it. Perhaps something you can consider down the road.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -424,3 +398,51 @@ def drain_queue(worker_type, dbt, event):
...
+async def consume(job_counter, worker_type, dbt):
+  """Retrieve celery results from redis db"""
+  redis = await aioredis.from_url(
+      f"redis://{TUNA_CELERY_BROKER}:{TUNA_REDIS_PORT}/15")
+  LOGGER.info('Connected to redis')
+  while job_counter.value > 0:
+    cursor = "0"
+    keys = []
+    while cursor != 0:
+      cursor, results = await redis.scan(cursor, match='*'
```
JehandadKhan: The match pattern needs to be updated to the prefix you set above.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -424,3 +398,51 @@ def drain_queue(worker_type, dbt, event):
...
+                                        )  # update with the celery pattern
+      keys.extend(results)
+    LOGGER.info('Found results')
```
JehandadKhan:
```suggestion
    LOGGER.info(f'Found {len(restults)} results')
```

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -424,3 +398,51 @@ def drain_queue(worker_type, dbt, event):
...
+    for key in keys:
+      LOGGER.info(type(key))
```
JehandadKhan: not necessary anymore.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -424,3 +398,51 @@ def drain_queue(worker_type, dbt, event):
...
+      try:
+        data = await redis.get(key)
+        if data:
+          await result_callback(key.decode('utf-8'), data.decode('utf-8'), worker_type, dbt)
```
JehandadKhan: You can have your callback work on batches of data items, to make data insertion in the database more effective.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -424,3 +398,51 @@ def drain_queue(worker_type, dbt, event):
...
+          await redis.delete(key)
+          job_counter.value - 1
+      except aioredis.exceptions.ResponseError as red_err:
+        LOGGER.error(red_err)
+        LOGGER.info(key.decode('utf-8'))
+
+    LOGGER.info('Processed results')
+    await asyncio.sleep(1)
```
JehandadKhan, line 444: For proof of concept/hacking this is ok, but I am not sure how I feel about it otherwise.

tuna/miopen/celery_tuning/tuning.py
```diff
@@ -424,3 +398,51 @@ def drain_queue(worker_type, dbt, event):
...
+    context = data['result']['context']
+    LOGGER.info('Parsing: %s', fin_json)
+    LOGGER.info('Parsing context: %s', context)
+    if worker_type == 'fin_build_worker':
+      process_fin_builder_results(session, fin_json, context, dbt)
+    elif worker_type == 'fin_eval_worker':
+      process_fin_evaluator_results(session, fin_json, context, dbt)
+    else:
+      raise ValueError('Unsupported worker_type')
+
```
JehandadKhan: Is this still required?

tuna/miopen/celery_tuning/tuning.py
alexandraBara: we dont have that support on the parsing side yet. But yes that is the plan

tuna/miopen/celery_tuning/tuning.py
alexandraBara: maybe once we have the ability to reset state on ctrl+c we can look into it.

tuna/miopen/celery_tuning/tuning.py
alexandraBara: if you mean the callback function, then yes still required, it does the whole parsing of results - it is technically not a callback function so its probably best to rename so we avoid confusion

tuna/miopen/celery_tuning/tuning.py
alexandraBara: i was just piggybacking off a function we already have

tuna/celery_app/celery_workers.py
```diff
@@ -0,0 +1,115 @@
...
+LOGGER: logging.Logger = setup_logger('celery_workers')
+
+
+def launch_worker_per_node(machines, cmd, formatted=False):
+  """Launch celery worker for compile"""
+  final_cmd = cmd
+  subp_list = []
+  for machine in machines:
+    try:
+      if formatted:
```
alexandraBara, line 43: @cderb  can we remove the formatted bool? there is no else part, what is the intention for that?

cderb: I didn't add this portion. replace should function correctly weather HOSTNAME is present or not.

alexandraBara:
Posting different branch with updated code

PR #964, opened 2024-04-11
Celery: queue purge on CTRL+C

alexandraBara:
adding CTRL+C handler for purging tasks that have been enqueued. Celery does a warm shutdown on the workers but the tasks left in redis do not get purged.

### DIFF
--- a/tuna/celery_app/celery_app.py
+++ b/tuna/celery_app/celery_app.py
@@ -26,6 +26,7 @@
 ###############################################################################
 """Module to define celery app"""
 import os
+from subprocess import Popen, PIPE
 from celery import Celery
 from celery.utils.log import get_task_logger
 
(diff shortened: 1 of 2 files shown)


tuna/celery_app/celery_app.py
```diff
@@ -55,5 +56,20 @@ def stop_active_workers():
   return True
 
 
+def purge_queue(q_names, logger):
+  """Purge jobs in queue"""
+  for q_name in q_names:
+    try:
+      logger.info('Purging Q %s', q_name)
+      cmd = f"celery -A tuna.celery_app.celery_app purge -f -Q {q_name}".split(
+          ' ')
```
msaudulhassan, line 65: is it possible for `q_name` to have a space character `' '`?

msaudulhassan (approved):
Just general comments:

Some of our previous tunings managed to choke out system memory. Is a `MemoryError` exception something that should also be handled?

PR #966, opened 2024-04-30
GPU mismatch fixed in Eval

msaudulhassan:
MITuna allows for evaluation to be run even if the underlying GPU doesn't match the GPU configured for the tuning run. This PR prohibts this behaviour

### DIFF
--- a/tuna/miopen/worker/fin_eval.py
+++ b/tuna/miopen/worker/fin_eval.py
@@ -297,6 +297,22 @@ class FinEvaluator(FinClass):
         return False
     return True
 
+  def check_env(self) -> bool:
+    """Interface function to check the miopen env version vs presumed miopen version"""
+    if super().check_env():
+      if self.dbt.session.arch != self.machine.arch:
+        raise ValueError(
+            f'session arch {self.dbt.session.arch} does not match machine arch\
+            {self.machine.arch}')
+      if self.dbt.session.num_cu != self.machine.num_cu:
+        raise ValueError(
+            f'session num_cu {self.dbt.session.num_cu} does not match machine num_cu\
+            {self.machine.num_cu}')
+    else:
+      return False
+
+    return True
+
   def step(self):
     """Function that defined the evaluator specific functionality which implies picking up jobs
     to benchmark and updating DB with evaluator specific state"""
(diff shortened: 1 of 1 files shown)


tuna/miopen/worker/fin_eval.py
```diff
@@ -297,6 +297,22 @@ def manage_queue(self):
...
+    """Check the GPU on the machine matches the GPU specified in session table"""
+    if super().check_env():
+      if self.dbt.session.arch != self.machine.arch:
+        raise ValueError(
+            f'session arch {self.dbt.session.arch} does not match machine arch\
+            {self.machine.arch}')
+      if self.dbt.session.num_cu != self.machine.num_cu:
+        raise ValueError(
+            f'session num_cu {self.dbt.session.num_cu} does not match machine num_cu\
+            {self.machine.num_cu}')
```
cderb: These exceptions here should return false as opposed to throwing.

alexandraBara:
```suggestion
    if super().check_env():
      if self.dbt.session.arch != self.machine.arch or  self.dbt.session.num_cu != self.machine.num_cu:
        self.logger.error('Session arch/num_cu does not match env arch/num_cu')
        return False
```
Something like that

tuna/miopen/worker/fin_eval.py
```diff
@@ -300,14 +300,13 @@ def manage_queue(self):
...
-        raise ValueError(
-            f'session num_cu {self.dbt.session.num_cu} does not match machine num_cu\
-            {self.machine.num_cu}')
+      if self.dbt.session.arch != self.machine.arch or \
+              self.dbt.session.num_cu != self.machine.num_cu:
+        self.logger.error(
+            'Session arch/num_cu (%s/%s) does not match env arch/num_cu (%s/%s)',
+            self.dbt.session.arch, self.dbt.session.num_cu, self.machine.arch,
+            self.machine.num_cu)
+        return False
```
alexandraBara, line 309: @cderb  would this be better moved in the fin_class check_env function? that way both eval and compile can check?

cderb: Compile is capable of running on other architectures.
Though applicability needs to be run on the same architecture because of CK.
So there is some complexity here.

cderb (approved):

alexandraBara (approved):

PR #969, opened 2024-06-06
Tuna Docker updates

cderb:
add ARCH_TARGET to dockerfile args to limit build for ck
change backed for .jenkins/update_solvers to HIP for applicability 
Dockerfile link /opt/rocm if rocm directory exists with extension 
Dockerfile update installation of miopen requirements 
Dockerfile build miopen with offline clang compiler session table increase docker string length to handle staging image names 
utils.groovy update build args to include machine arch

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,14 @@ FROM $USEIMAGE as dtuna-ver-0
 #args before from are wiped
 ARG ROCMVERSION=
 ARG OSDB_BKC_VERSION=
-ARG BUILD_MIOPEN_DEPS=
+
+RUN test -d /opt/rocm*; \
+    if [ $? -eq 0 ] ; then \
+        test -d /opt/rocm; \
+        if [ $? ] ; then \
+            ln -s /opt/rocm* /opt/rocm; \
+        fi \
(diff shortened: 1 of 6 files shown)


alexandraBara (approved):

Dockerfile
```diff
@@ -6,7 +6,7 @@ ARG OSDB_BKC_VERSION=
 ARG HASVER=${ROCMVERSION:+$ROCMVERSION}
 ARG HASVER=${HASVER:-$OSDB_BKC_VERSION}
 
-ARG BASEIMAGE=rocm/miopen:ci_c1ca2a
+ARG BASEIMAGE=rocm/miopen:ci_5fec03
```
reidkwja, line 9: Can this be passed as an environmental variable + docker run arg instead?

reidkwja (changes requested):

Dockerfile
cderb: BASEIMAGE is passable as a docker-arg

reidkwja (approved):

msaudulhassan (approved):

PR #972, opened 2024-07-12
Celery utest coverage

alexandraBara:


reidkwja (approved):

tests/test_celery.py
```diff
@@ -0,0 +1,277 @@
...
+
+  #load jobs
+  dbt = MIOpenDBTables(config_type=ConfigType.convolution)
+  num_jobs = add_test_jobs(miopen, miopen.args.session_id, dbt,
+                           miopen.args.label, miopen.args.label,
+                           ['miopen_perf_compile'],
+                           'test_add_celery_compile_job',
+                           'miopenConvolutionAlgoGEMM')
+  #assert num_jobs
+  num_jobs = 4
```
cderb, line 65: comment is assert, action is set

tests/test_celery.py
```diff
@@ -0,0 +1,277 @@
...
+
+  cmd = f"celery -A tuna.celery_app.celery_app worker -l info -E -n tuna_HOSTNAME_sess_{miopen.args.session_id} -Q test_{db_name}"  #pylint: disable=line-too-long
+  #testing launch_worker_per_node
+  subp_list = launch_worker_per_node([machine], cmd, True)
+  sleep(2)
+  assert subp_list
+  assert miopen.cancel_consumer(f"test_{db_name}")
+  for subp in subp_list:
+    print(subp.pid)
+    subp.kill()
```
cderb: cancel_consumer would also kill the workers for the named queue. Perhaps this loop would confirm that the subp were terminated by this call.

tests/test_celery.py
alexandraBara: leftover from testing, ill remove

tests/test_celery.py
alexandraBara: i'll add

cderb (approved):

PR #980, opened 2024-08-27
Celery: Example update

alexandraBara:
Updated Example project to use celery
Updated Example documentation
Moved more functionality into mituna_interface 
Updated code cov for example


tuna/example/doc/Tuning.md
```diff
@@ -33,25 +33,46 @@ The next command is:
...
 
-The third step is:
+Setting up jobs for tuning:
 
 .. code-block::  
 
   ./tuna/example/load_job.py -a gfx908 -n 120 -l my_label --session_id 1
 
-This steps loads jobs in the *job* table. These jobs will be picked up for execution in the
+This steps loads jobs in the *job* table. ThNoneese jobs will be picked up for execution in the
```
cderb: typo

tuna/utils/celery_utils.py
```diff
@@ -0,0 +1,45 @@
...
+#
+###############################################################################
+"""Utility module for celery helper functions"""
+from tuna.utils.utility import SimpleDict
+
+
+def prep_default_kwargs(kwargs, args, cached_machine):
+  """Populate kwargs with serialized job and machine"""
+  kwargs["job"] = SimpleDict(**args[0])
+  kwargs["machine"] = cached_machine
```
cderb: The context for cached_machine is lost here. Just machine would be sufficient.
Similarly the transformation on args is oddly specific for this function (the context is lost). I'd suggest the parameters to this function would be kwargs, job, machine.

cderb (approved):

tuna/example/celery_tuning/celery_tasks.py
```diff
@@ -0,0 +1,76 @@
...
+
+@celeryd_after_setup.connect
+def capture_worker_name(sender, instance, **kwargs):  #pylint: disable=unused-argument
+  """Capture worker name"""
+  app.worker_name = sender
+
+
+cached_machine = Machine(local_machine=True)
+
+
```
msaudulhassan, line 49: It'd be nice to do away with these global variables

tuna/example/example_lib.py
```diff
@@ -166,26 +196,81 @@ def get_kwargs(self,
...
 
   def celery_enqueue_call(self, context, q_name, task_id=False):
-    """Wrapper function for celery enqueue func"""
-    raise NotImplementedError('Not implemented')
+    """! Wrapper function for celery enqueue func
+    @param context serialized context for Celery job
+    @param q_name Name of custom Celery queue
+    @param task_id custom task ID for redis Key
+    """
+    Q_NAME = q_name  #pylint: disable=import-outside-toplevel,unused-variable,invalid-name,redefined-outer-name
```
msaudulhassan, line 245: I see there's a global variable named `Q_NAME` as well. Is this supposed to be a reference to that variable (in which case you'd need `global Q_NAME` before this assignment)? If not, what's the use for the global `Q_NAME`?

alexandraBara: this is a side effect of needing the global Q_NAME at import time but not being able to assign it a value until runtime. Since imports are loaded before the program runs this workaround has to be done. Q_NAME here is a class variable and is visible where this assignment is made, so we dont need the global keyword here. We were OK with this design (although not ideal) because the other option would be to define it in an env variable and we did not want to add another env var for this.

tuna/example/celery_tuning/celery_tasks.py
alexandraBara: We did not have one in the beginning, but a Machine is not a serializable object and hence we cannot pass it to a celery worker and must be constructed for every celery task. This 'cached_machine' variable acts as a cache so we dont have to create a new Machine object if one has already been created. Its simply a optimization addition, but yes it does add a global variable.

msaudulhassan (approved):

PR #984, opened 2024-10-18
RabbitMQ connection heartbeat

alexandraBara:
rabbitMQ fails to close stale connections by default and we run out of sockets (available connection).
Added heartbeat check at 60sec to close unresponsive connections and free up space.


tuna/celery_app/celery_app.py
```diff
@@ -87,6 +87,7 @@ def get_backend_env():
     f"amqp://{TUNA_CELERY_BROKER_USER}:{TUNA_CELERY_BROKER_PWD}@{TUNA_CELERY_BROKER_HOST}:{TUNA_CELERY_BROKER_PORT}/",
     result_backend=
     f"redis://{TUNA_CELERY_BACKEND_HOST}:{TUNA_CELERY_BACKEND_PORT}/15",
+    broker_transport_options={"heartbeat": 60},
```
reidkwja, line 90: It might be a good idea to make that an env too? saves additional PR if the interval needs to be adjusted in future...

cderb (approved):

reidkwja (approved):

tuna/celery_app/celery_app.py
alexandraBara: i dont see us changing this too often, but if we do, I will add an env

PR #986, opened 2024-11-05
update export_db script to export pdb in db.txt format

cderb:


tuna/miopen/subcmd/export_db.py
```diff
@@ -89,13 +93,62 @@ def get_filename(arch: str,
...
   else:
-    extension = ".db"
+    extension = ".db.txt"
 
   final_name = f"{final_name}{extension}"
 
   return final_name
 
 
+def fin_net_cfg_job(cfg_lst):
```
alexandraBara: should pass in config_type as well and can put in place "Not implemented " for non-conv case. It matters when u build the job and add the "direction" which is conv specific hardcoded. I would say throughout any new function you add here that has elements specific to conv, add the case for non-conv and throw ValueError() or something like that so we can leave a trace to update later.

alexandraBara (changes requested):

tuna/miopen/subcmd/export_db.py
```diff
@@ -89,13 +94,67 @@ def get_filename(arch: str,
...
+                    config_type: ConfigType = ConfigType.convolution):
+  """Construct a fin network_config job from a config
+  """
+  #arch and num_cu are required by fin, but unused for this command
+  job_list = []
+  if config_type == ConfigType.convolution:
+    for config in cfg_lst:
+      job = {
+          "steps": ["network_config"],
+          "arch": 'gfx908',
```
alexandraBara, line 115: should probably not be hardcoded here

tuna/miopen/subcmd/export_db.py
```diff
@@ -89,13 +94,67 @@ def get_filename(arch: str,
...
+          "num_cu": 120,
+          "config_tuna_id": config.id,
+          "direction": int(INVERS_DIR_MAP[config.direction]),
+          "config": compose_config_obj(config)
+      }
+      job_list.append(job)
+  else:
+    logger.error(f"Config type not implemented: {config_type}")
```
alexandraBara, line 123: nice!

tuna/miopen/subcmd/export_db.py
```diff
@@ -406,6 +465,27 @@ def insert_perf_db_sqlite(cnx, perf_db_entry, ins_cfg_id):
   return perf_db_dict
 
 
+def populate_sqlite(cfg_map, num_perf, cnx, perf_db_entry, cfg_entry,
+                    total_entries, logger: logging.Logger):
+  """Analyze perf_dv entry"""
```
alexandraBara: spell check

tuna/miopen/subcmd/export_db.py
```diff
@@ -428,25 +508,65 @@ def export_pdb(dbt: MIOpenDBTables, args: argparse.Namespace,
...
+      lst = perf_db.get(db_key)
+      if not lst:
+        perf_db[db_key] = [pdb_entry]
+      else:
+        lst.append(pdb_entry)
+
+  return perf_db
+
+
+def write_pdb(arch, num_cu, ocl, perf_db, filename=None):
```
alexandraBara, line 543: probably not a good idea to default filename to None it seems like it needs to be not none based on your code

alexandraBara (changes requested):

reidkwja (approved):

tuna/miopen/subcmd/export_db.py
cderb: Comment on line 109, fin expects a valid arch/cu in the fin command, but the network_config step will not use it.

tuna/miopen/subcmd/export_db.py
cderb: get_filename would generate an appropriate name if filename is None.

tuna/miopen/subcmd/export_db.py
cderb: fixed

alexandraBara (approved):

PR #996, opened 2025-06-25
Add attn split kv

dhernandez0:
Merge this after https://github.com/ROCm/rocMLIR/pull/1895 is merged


dorde-antic (changes requested):
Should we add split_kv to line 742?
`    for datatype, transQ, transK, transV, transO, withAttnScale, withAttnBias, causal, return_lse, line in \
              itertools.product(['f32', 'f16'], ['false', 'true'],
                                ['false', 'true'], ['false', 'true'],
                                ['false', 'true'], ['false', 'true'], ['false', 'true'], ['false', 'true'], ['false', 'true'], lines):
        line = line.strip()`

and then append to one_config after line 763?

dorde-antic (approved):

umangyadav (approved):
We should test this branch in CI before merging this. 
You can change MITuna branch that is being used in Jenkinsfile to point to this branch.

cderb (approved):

PR #999, opened 2025-07-18
Propagate errors correctly, remove unreachable part of code

dorde-antic:
This PR ensures that errors are correctly propagated by ensuring that failures during subprocess calls are correctly handled. The goal of PR is to avoid silent errors in rocMLIR's CI runs.

### DIFF
--- a/tuna/worker_interface.py
+++ b/tuna/worker_interface.py
@@ -27,6 +27,7 @@
 """Module that represents the WorkerInterface class interface"""
 
 from multiprocessing import Process, Lock
+import sys
 try:
   import queue
 except ImportError:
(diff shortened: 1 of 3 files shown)


tuna/worker_interface.py
```diff
@@ -518,11 +521,6 @@ def run(self) -> bool:  #type: ignore[override]
         self.num_procs.value -= 1
       return False
 
-    with self.bar_lock:
```
umangyadav, line 521: I am not sure what this does. but why was it removed ?

dorde-antic: If I'm not wrong that part of code is unreachable

PR #1005, opened 2025-08-15
Collect comprehensive tuning results

cderb:
## Motivation

Enable collection + storage of multiple perf parameter results per config + solver

## Technical Details

Added additional table ConvolutionTuningData to function like ConvolutionFindDB except with params as an additional condition for uniqueness. 
Added update methods to store fin results to this table.
Added --rich_data as an option to be ran with go_fish enqueue to store results to the new table.

## Test

Ran on 50 configurations with Tunable and Non-Tunable solvers. Population of the ConvolutionFindDB table was unaffected. Population of the ConvolutionTuningData table was successful for both types of solvers.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,5 @@
 [run]
+relative_files = True
 source = ./tuna
 omit = 
       /usr/local/lib* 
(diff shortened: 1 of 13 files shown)


reidkwja (approved):
lgtm

